Issue #17, opened 2017-07-19
Knack doesn't support validators


tjprescott:
This is a crucial feature that Azure CLI relies upon, is bizarrely not supported by argparse, and really should be part of Knack.  We should include it.  Currently I have to monkey patch CommandInvoker.execute to reintroduce this functionality.

...

However, while adding this feature to Knack, please, please, P-L-E-A-S-E including both of the following:

1. Argument level validators (the kind currently supported in AzureCLI)
2. Command level validators (NOT supported in AzureCLI)

If a command has a validator, then any argument level validators that would ordinarily be applied due to the argument registry aliasing, blah blah, would be ignored. A command can have at most ONE command validator or many argument level validators.  The single command-level validator would allow you to sequence multiple validators, which is often necessary.  The current way, of attaching a validator to a random property and making sure to strip any other ones, falls under the category of hacky workaround so we shouldn't give knack this handicap.

PR #21, opened 2017-08-14
Support command-level and argument-level validators

derekbekoe:
Closes https://github.com/Microsoft/knack/issues/17

### DIFF
--- a/docs/validators.md
+++ b/docs/validators.md
@@ -10,9 +10,11 @@ If a command has a command-level validator, then any argument-level validators t
 i.e. A command can have at most command validator or many argument level validators.
 
 **Command-level Validators**
+
 Command-level validators can operate on any arguments on the command. This is useful when you want a validator to make use of more than a single argument.
 
 **Argument-level Validators**
+
 Argument-level validators should only operate on a single argument.
 
 The order argument-level validators are executed in is not guaranteed so don't use an arguments that make use of multiple arguments.
--- a/docs/validators.md
+++ b/docs/validators.md
@@ -1,0 +1,19 @@
+Validators
+==========
+
+Validators allow you to validate or transform command arguments just before execution.
+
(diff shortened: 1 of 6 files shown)


docs/validators.md
```diff
@@ -0,0 +1,21 @@
...
+
+Knack supports both command-level and argument-level validators.
+
+If a command has a command-level validator, then any argument-level validators that would ordinarily be applied are ignored.
+
+i.e. A command can have at most command validator or many argument level validators.
+
+**Command-level Validators**
+
+Command-level validators can operate on any arguments on the command. This is useful when you want a validator to make use of more than a single argument.
```
tjprescott: An argument validator can always use multiple arguments. I think a more accurate statement here would be "This is useful when you need to control the validation sequence."

The best example I can give is when you want to use the "get_location_from_resource_group" validator. This would require a command validator because the location needs to be resolved first if other validators might need to make use of the computed location.

docs/validators.md
```diff
@@ -0,0 +1,21 @@
...
+
+**Argument-level Validators**
+
+Argument-level validators should only operate on a single argument.
+
+The order argument-level validators are executed in is not guaranteed so don't use an arguments that make use of multiple arguments.
```
tjprescott: Yeah, this isn't excactly true. For example, you could have a single validator that makes use of `subnet_name` and `vnet_name` for example.  It's totally fine to have this in a single validator, just as long as some OTHER validator doesn't rely on them.

docs/validators.md
derekbekoe: okay thanks done.

docs/validators.md
derekbekoe: okay I've tried to clarify this...

derekbekoe:
@tjprescott Feedback addressed. Can you take another look at the doc changes?

Issue #11, opened 2017-07-10
Arguments _NAMED_ARGUMENTS cannot be changed


tjprescott:
This variable, set in CLICommandArgument, is a tuple. As such, I cannot subclass CLICommandArgument and change the value. I need to add 'id_part' to the list, but due to the immutable nature of Python tuples, I cannot do this. If this were changed to a simple list, there wouldn't be an issue.

tjprescott:
Currently this has me blocked. I would need to find a different way to implement `--ids` in the CLI.

tjprescott:
Additionally I'm running into similar issues with `configured_default`. It fails at:
https://github.com/Microsoft/knack/blob/master/knack/parser.py#L74
Essentially it is trying to create an argparse argument with unrecognized kwargs.

It is very, VERY likely that someone would want to support kwargs above and beyond the ones argparse does (obviously, as that's what we're doing!).  We should have a more consistent way of separating out the argparse kwargs from custom kwargs.  Right now we tend to try and process our custom kwargs early and then pop them off so that argparse is happy, but there should really be a more methodical approach.

tjprescott:
Here's the workaround I used in az, which should really be in knack:

```Python
def _add_argument(self, obj, arg):
        options_list = arg.options_list
        argparse_options = {name: value for name, value in arg.options.items() if name in ARGPARSE_SUPPORTED_KWARGS}
        return obj.add_argument(*options_list, **argparse_options)
```

Unfortunately, I had to just copy-paste the list of supported kwargs from argsparse since it didn't expose them as a property or method.  Still, this is CLI neutral and more reliable that having to remember to pop non-compliant kwargs.

PR #25, opened 2017-09-26
Allow overriding of NAMED_ARGUMENTS

derekbekoe:
Closes https://github.com/Microsoft/knack/issues/11

### DIFF
--- a/knack/parser.py
+++ b/knack/parser.py
@@ -23,6 +23,7 @@ ARGPARSE_SUPPORTED_KWARGS = [
     'required',
     'help',
     'metavar',
+    'action'
 ]
 
 
--- a/knack/parser.py
+++ b/knack/parser.py
@@ -10,6 +10,22 @@ from .help import show_help
 from .util import CtxTypeError
 
 
+# List of keyword arguments supported in argparse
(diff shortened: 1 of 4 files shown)


knack/parser.py
```diff
@@ -10,6 +10,22 @@
...
+    'dest',
+    'nargs',
+    'const',
+    'default',
+    'type',
+    'choices',
+    'required',
+    'help',
+    'metavar',
+]
```
derekbekoe: In your 'KnackConversion' branch, [you also included 'action'](https://github.com/Azure/azure-cli/blob/KnackConversion/src/azure-cli-core/azure/cli/core/parser.py#L21) but I haven't included it in the PR to Knack as I didn't see it in argparse in https://github.com/python/cpython/blob/master/Lib/argparse.py#L748.

derekbekoe: Is this good or is there a scenario where 'action' should be added?

tjprescott: Action is definitely supported by argparse and thus should be in the list.

tjprescott: `action='store_true'`  is an example.

derekbekoe:
@tjprescott I've added 'action' back in

Issue #10, opened 2017-07-07
Eliminate redundant ways of doing the same thing


tjprescott:
Currently, knack allows you to register commands with either of these options:

```
# method 1
self.cli_command(...)

# method 2
with CommandSuperGroup(...) as sg:
   with CommandGroup(...) as g:
     g.command(...)
```

Argument registration likewise exposes two syntactically very different ways of doing the same thing. Neither method is clearly better--they are both just different.  However, this has resulted in confusion for external partners as to what the difference is and what they should use.  While it doesn't take hours to tell them "they are the same and use whichever you prefer", ideally there would only be one right way to do things, and by this point we should have a better idea what that is (my guess is it would be a third implementation that combines the strengths and mitigates the weaknesses of both existing approaches).

Since this is supposed to be a framework that exposes the simplest way to do the right thing for external partners, we should strive to commit to only a single way of registering commands and arguments, especially give that no implementation can be migrated to the new framework without modification.

tjprescott:
Also, note that CLI 2.0 already subclasses CLICommandsLoader, so even if we wanted to preserve the "old ways" we can do that there. There's no need to enshrine that in Knack (on the contrary, there's great reasons NOT to!)

PR #26, opened 2017-09-27
Eliminate redundant ways of doing the same thing

derekbekoe:
Closes https://github.com/Microsoft/knack/issues/10


knack/commands.py
```diff
@@ -173,16 +169,6 @@ def user_confirmed(confirmation, command_args):
...
-        self.argument_registry.register_cli_argument(scope, dest, arg_type, **kwargs)
-
-    def register_extra_cli_argument(self, command, dest, **kwargs):
-        '''Register extra parameters for the given command. Typically used to augment auto-command built
-        commands to add more parameters than the specific SDK method introspected.
-        '''
-        self.extra_argument_registry[command][dest] = CLICommandArgument(dest, **kwargs)
-
 
 class CommandSuperGroup(object):
```
tjprescott: In my experience converting Az, CommandSuperGroup has no use.

derekbekoe: SF makes use of this - https://github.com/Azure/service-fabric-cli/blob/master/src/sfctl/commands.py#L40.
Also, it's used a few times in az as `ServiceGroup`.

Issue #8, opened 2017-07-07
Documentation deficiencies


tjprescott:
Most public facing methods and constructors lack docstrings that explains what the various kwargs do. This makes it very hard to use anything other than those described in the examples, which are very simple.

PR #34, opened 2017-09-29
Better documentation with docstrings

derekbekoe:
Closes https://github.com/Microsoft/knack/issues/8

### DIFF
--- a/knack/arguments.py
+++ b/knack/arguments.py
@@ -129,7 +129,8 @@ class ArgumentsContext(object):
 
         :param command_loader: The command loader that arguments should be registered into
         :type command_loader: knack.commands.CLICommandsLoader
-        :param command_scope: The command scope this context manager should have
+        :param command_scope: The scope to which arguments in this context apply.
+                              More specific scopes will override less specific scopes in the event of a conflict.
         :type command_scope: str
         """
         self.command_loader = command_loader
@@ -144,7 +145,7 @@ class ArgumentsContext(object):
     def argument(self, argument_dest, arg_type=None, **kwargs):
         """ Register an argument for the given command scope using a knack.arguments.CLIArgumentType
 
-        :param argument_dest: The destination argument to add this argument type to
+        :param argument_dest: Predefined CLIArgumentType definition to register, as modified by any provided kwargs.
         :type argument_dest: str
         :param arg_type: The argument type we are registering
         :type arg_type: knack.arguments.CLIArgumentType
@@ -154,18 +155,6 @@ class ArgumentsContext(object):
                                                                     argument_dest,
                                                                     arg_type,
                                                                     **kwargs)
-
-    def register(self, argument_dest, options_list, **kwargs):
-        """ Register an argument for the given command scope without a knack.arguments.CLIArgumentType.
-            You add the kwargs you require yourself.
-
-        :param argument_dest: The destination argument to add this argument type to
-        :type argument_dest: str
-        :param options_list: The options that can be supplied for this argument
-        :type options_list: tuple
-        :param kwargs: Any argument type overrides that need to be applied
-        """
-        self.argument(argument_dest, arg_type=None, options_list=options_list, **kwargs)
 
     def ignore(self, argument_dest):
         """ Register an argument with type knack.arguments.ignore_type (hidden/ignored)
--- a/knack/arguments.py
+++ b/knack/arguments.py
@@ -12,9 +12,16 @@ logger = get_logger(__name__)
 
 
 class CLIArgumentType(object):
+
     REMOVE = '---REMOVE---'
 
     def __init__(self, overrides=None, **kwargs):
+        """A base CLI Argument Type that can be applied to multiple command arguments
+
+        :param overrides: The base argument that you are overriding
+        :type overrides: knack.arguments.CLIArgumentType
+        :param kwargs: Kwargs for an argument (Argparse keywords can be used or you can add your own)
+        """
         if isinstance(overrides, str):
             raise ValueError("Overrides has to be a {} (cannot be a string)".format(CLIArgumentType.__name__))
         options_list = kwargs.get('options_list', None)
@@ -30,9 +37,18 @@ class CLIArgumentType(object):
 
 
 class CLICommandArgument(object):  # pylint: disable=too-few-public-methods
+
     NAMED_ARGUMENTS = ['options_list', 'validator', 'completer', 'arg_group']
 
     def __init__(self, dest=None, argtype=None, **kwargs):
+        """An argument that has a specific destination parameter.
+
+        :param dest: The parameter that this argument is for
+        :type dest: str
+        :param argtype: The argument type for this command argument
+        :type argtype: knack.arguments.CLIArgumentType
+        :param kwargs: see knack.arguments.CLIArgumentType
+        """
         self.type = CLIArgumentType(overrides=argtype, **kwargs)
         if dest:
             self.type.update(dest=dest)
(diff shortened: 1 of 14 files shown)


knack/arguments.py
```diff
@@ -12,9 +12,16 @@
...
 class CLIArgumentType(object):
+
     REMOVE = '---REMOVE---'
 
     def __init__(self, overrides=None, **kwargs):
+        """A base CLI Argument Type that can be applied to multiple command arguments
+
+        :param overrides: The base argument that you are overriding
+        :type overrides: knack.arguments.CLIArgumentType
+        :param kwargs: Kwargs for an argument (Argparse keywords can be used or you can add your own)
```
tjprescott: There are kwargs that Knack supports beyond the argparse keywords (validator, for example). These should be called out here, especially since everything else points to this (however, I don't think it should point here).

knack/arguments.py
```diff
@@ -87,63 +124,82 @@ def get_cli_argument(self, command, name):
 
 
 class ArgumentsContext(object):
-    def __init__(self, command_loader, command):
+    def __init__(self, command_loader, command_scope):
+        """ Context manager to register arguments
+
+        :param command_loader: The command loader that arguments should be registered into
+        :type command_loader: knack.commands.CLICommandsLoader
+        :param command_scope: The command scope this context manager should have
```
tjprescott: Rec: "The scope to which arguments in this context apply. More specific scopes will override less specific scopes in the event of a conflict."

knack/arguments.py
```diff
@@ -87,63 +124,82 @@ def get_cli_argument(self, command, name):
...
-
-    def argument(self, argument_name, arg_type=None, **kwargs):
-        self.command_loader.argument_registry.register_cli_argument(self.commmand,
-                                                                    argument_name,
+    def argument(self, argument_dest, arg_type=None, **kwargs):
+        """ Register an argument for the given command scope using a knack.arguments.CLIArgumentType
+
+        :param argument_dest: The destination argument to add this argument type to
+        :type argument_dest: str
+        :param arg_type: The argument type we are registering
```
tjprescott: This makes it sound like you have to register an arg_type, but you really don't (and often don't in fact). Rec: "Predefined CLIArgumentType definition to register, as modified by any provided kwargs."

knack/arguments.py
```diff
@@ -87,63 +124,82 @@ def get_cli_argument(self, command, name):
...
+        :type arg_type: knack.arguments.CLIArgumentType
+        :param kwargs: Any argument type overrides that need to be applied
```
tjprescott: IMO, *this* is where you should call out exactly which kwargs are natively supported by Knack, like "validator", "confirmation", etc. This is the method most people will use to register their arguments, so this is the thing that, more than the other methods that would rarely be used (and arguably shouldn't be used), would need a thorough docstring to provide useful intellisense.

knack/arguments.py
```diff
@@ -87,63 +124,82 @@ def get_cli_argument(self, command, name):
...
                                                                     arg_type,
                                                                     **kwargs)
 
-    def register_alias(self, argument_name, options_list, **kwargs):
-        self.command_loader.argument_registry.register_cli_argument(self.commmand,
-                                                                    argument_name,
-                                                                    None,
-                                                                    options_list=options_list,
-                                                                    **kwargs)
+    def register(self, argument_dest, options_list, **kwargs):
```
tjprescott: Arguably, this isn't needed at all since the `arg_type` in `argument` is optional. Also, an option list shouldn't be needed since `argument` applies the heuristic to the `dest` value.  I would (and do) use argument consistently, whether I supply an `arg_type` or not.

knack/arguments.py
```diff
@@ -87,63 +124,82 @@ def get_cli_argument(self, command, name):
...
+        """ Register an argument with type knack.arguments.ignore_type (hidden/ignored)
+
+        :param argument_dest: The destination argument to apply the ignore type to
+        :type argument_dest: str
+        """
+        self.argument(argument_dest, arg_type=ignore_type)
 
     def extra(self, dest, **kwargs):
         '''Register extra parameters for the given command. Typically used to augment auto-command built
         commands to add more parameters than the specific SDK method introspected.
```
tjprescott: I'd recommend a copy-paste of the kwargs here as well since this is an end-point that the user would likely want Intellisense for.

knack/commands.py
```diff
@@ -27,6 +27,29 @@ class CLICommand(object):  # pylint:disable=too-many-instance-attributes
...
+        :type description: str
+        :param table_transformer: A function that transforms the command output for displaying in a table
+        :type table_transformer: function
+        :param arguments_loader: The function that defines how the arguments for the command should be loaded
+        :type arguments_loader: function
+        :param description_loader: The function that defines how the description for the command should be loaded
+        :type description_loader: function
+        :param formatter_class: The formatter for how help should be displayed
+        :type formatter_class: class
+        :param deprecate_info: Command deprecation information
```
tjprescott: I would recommend a bit more detail here. I'd recommend this, assuming it is technically accurate: "Deprecation message to display when this command is invoked."

knack/commands.py
```diff
@@ -184,13 +242,33 @@ def __exit__(self, exc_type, exc_val, exc_tb):
         pass
 
     def group(self, group_name, **kwargs):
+        """ Creates a new group of commands that inherits from the super group.
+
+        :param group_name: The name of this group of commands in the command hierarchy
+        :type group_name: str
+        :param kwargs: Kwargs to apply to all commands in this group
```
tjprescott: I would call out the supported kwargs here.

knack/commands.py
```diff
@@ -203,12 +281,20 @@ def __enter__(self):
...
 
-    def command(self, name, method_name, **kwargs):
+    def command(self, name, handler_name, **kwargs):
+        """ Register a command into the command table
+
+        :param name: The name of the command
+        :type name: str
+        :param handler_name: The name of the handler that will be applied to the operations template
+        :type handler_name: str
+        :param kwargs: Kwargs to apply to the command
```
tjprescott: Call out the kwargs here.

knack/arguments.py
derekbekoe: done.

knack/commands.py
derekbekoe: done.

knack/arguments.py
derekbekoe: done.

derekbekoe:
I do call out the kwargs available in https://github.com/Microsoft/knack/blob/master/docs/arguments.md but I'll include them in code for intellisense for VS/IDE users.

knack/arguments.py
derekbekoe: Agreed. Removed.

derekbekoe:
What about pointing users to the doc page from the docstring instead of repeatedly calling out the kwargs everywhere?

tjprescott:
@derekbekoe the main purpose of the docstrings (in my head when I created the issue) was for intellisense. Pointing someone to the docs is okay (assuming the docs are correct) but as a command author I think I would much rather see it in the intellisense.  I know I personally found it maddening when `resource_id` switched to kwargs and I had to have the code (or docs) open side-by-side to know what to put. It was made much less annoying when the kwargs were listed in the Intellisense.

knack/arguments.py
derekbekoe: done

knack/arguments.py
derekbekoe: done.

derekbekoe:
@tjprescott How does my latest commit change look?

tjprescott:
@derekbekoe looks good to me :)

PR #47, opened 2017-11-15
Performance optimization

troydai:
1. Delay import platform
2. Delay import colorama

### DIFF
--- a/knack/log.py
+++ b/knack/log.py
@@ -4,11 +4,8 @@
 # --------------------------------------------------------------------------------------------
 
 import os
-import platform
 import logging
 from logging.handlers import RotatingFileHandler
-
-import colorama
 
 from .util import CtxTypeError, ensure_dir
 from .events import EVENT_INVOKER_PRE_CMD_TBL_CREATE, EVENT_PARSER_GLOBAL_CREATE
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -9,7 +9,7 @@
 from codecs import open
 from setuptools import setup, find_packages
 
-VERSION = '0.2.0'
+VERSION = '0.2.1'
```
derekbekoe: No need to bump version. We create a release commit at release time.

knack/log.py
```diff
@@ -35,19 +32,27 @@ def get_logger(module_name=None):
...
-        return wrap_msg_with_color
-
-    COLOR_MAP = {
-        logging.CRITICAL: _color_wrapper(colorama.Fore.RED),
-        logging.ERROR: _color_wrapper(colorama.Fore.RED),
-        logging.WARNING: _color_wrapper(colorama.Fore.YELLOW),
-        logging.INFO: _color_wrapper(colorama.Fore.GREEN),
-        logging.DEBUG: _color_wrapper(colorama.Fore.CYAN)
-    }
+        return cls.COLOR_MAP.get(level, None)
```
derekbekoe: Can you add a test to `test_log.py` for this change?
As more and more products depend on knack, I want to make sure the test coverage is there for changes.

troydai: ok.

troydai:
@derekbekoe test added and rebased.

PR #54, opened 2017-11-21
Add Tox support

troydai:
[Tox](https://tox.readthedocs.io/en/latest/) assist the project to automate under different python configurations.


derekbekoe:
Does `pytest` do the same thing as `python -m unittest discover tests`?
If so, what's the benefit of having it?

I also want it to do a `python setup.py bdist_wheel` of the package to check that it can be built correctly. It looks like that it doesn't happen with this change.

troydai:
The tox will build the package first and then install it before testing. `Pytest` is the latest test runner. `Nosetest` is not being actively developed. `Pytest` offers more execution option and better output then `unittest` package.

PR #66, opened 2018-01-03
[minor breaking change] Consolidate exception handling so tests can capture the right exception

yugangw-msft:
It is one of the options to address test breaks tracked by https://github.com/Azure/azure-cli/issues/5032
The other option is to have [test sdk](https://github.com/Azure/azure-cli/blob/dev/src/azure-cli-testsdk/azure/cli/testsdk/patches.py#L37) workaround by adding a new patch, but we still need an  extraction here say, to a method, so tests can mock.

//CC @troydai  @tjprescott  @derekbekoe

### DIFF
--- a/knack/cli.py
+++ b/knack/cli.py
@@ -157,8 +157,11 @@ class CLI(object):  # pylint: disable=too-many-instance-attributes
             func(self, **kwargs)
 
     def exception_handler(self, ex):  # pylint: disable=no-self-use
-        """ The default exception handler for unknown CLI exceptions. """
-        logger.exception(ex)
+        """ The default exception handler"""
+        if isinstance(ex, CLIError):
+            logger.error(ex)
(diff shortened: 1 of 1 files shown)


knack/cli.py
```diff
@@ -157,8 +157,11 @@ def raise_event(self, event_name, **kwargs):
             func(self, **kwargs)
 
     def exception_handler(self, ex):  # pylint: disable=no-self-use
-        """ The default exception handler for unknown CLI exceptions. """
-        logger.exception(ex)
+        """ The default exception handler"""
```
derekbekoe: nit: can you add a space after `handler`?

troydai:
Made change for ya. I will merge it when the CI passes.

troydai:
@derekbekoe do you approve this?

PR #72, opened 2018-02-05
Add CommandInvoker.OnPreTruncateCommandTable event

chewong:
Move 'CommandInvoker.OnPreTruncateCommandTable' event from azure-cli-core to knack https://github.com/Azure/azure-cli/pull/5460


tjprescott:
Knack does not do any command table truncating (AFAIK), so I would argue this change is not needed in Knack.

derekbekoe:
Agreed with @tjprescott. I was wondering the same.

chewong:
Talked with @troydai, decided not to make this a knack event.

PR #84, opened 2018-05-16
Add deprecation mechanism

### LINKED ISSUE #53: CLIHelp support command_help_file_cls
Currently, you have to override show_help for this and that shouldn't be required.

### LINKED ISSUE #79: Add test for docstring splitting
This functionality was carried over from CLI 2.0 but not the test.  It ensures that the docstring
```
""" This is line one. This is line two. """
```

Gets parsed by the introspection methods into:
```
short-summary: This is line one.
long-summary: This is line two.
```

tjprescott:
Fixes #53. Closes #79.

Allows user to supply different help classes than the default when creating a CLIHelp object.

Adds support for command, command group, argument, and argument option deprecation.

Improves support for Unicode characters when displaying help. This will likely require the following magic comment as the first line of the help file:

```Python
# coding=utf-8
```

See: https://github.com/Azure/azure-cli/issues/5954

### DIFF
--- a/knack/arguments.py
+++ b/knack/arguments.py
@@ -6,6 +6,7 @@
 import argparse
 from collections import defaultdict
 
+from .deprecation import Deprecated
 from .log import get_logger
 
 logger = get_logger(__name__)
@@ -39,7 +40,7 @@ class CLIArgumentType(object):
 
 class CLICommandArgument(object):  # pylint: disable=too-few-public-methods
 
-    NAMED_ARGUMENTS = ['options_list', 'validator', 'completer', 'arg_group']
+    NAMED_ARGUMENTS = ['options_list', 'validator', 'completer', 'arg_group', 'deprecate_info']
 
     def __init__(self, dest=None, argtype=None, **kwargs):
         """An argument that has a specific destination parameter.
@@ -100,8 +101,7 @@ class ArgumentRegistry(object):
         :type argtype: knack.arguments.CLIArgumentType
         :param kwargs: see knack.arguments.CLIArgumentType
         """
-        argument = CLIArgumentType(overrides=argtype,
-                                   **kwargs)
+        argument = CLIArgumentType(overrides=argtype, **kwargs)
         self.arguments[scope][dest] = argument
 
     def get_cli_argument(self, command, name):
@@ -143,6 +143,81 @@ class ArgumentsContext(object):
     def __exit__(self, exc_type, exc_val, exc_tb):
         pass
 
+    def _get_arg_deprecation_action(self, kwargs, deprecate_info):
+
+        # wrap any existing action
+        action = kwargs.get('action', None)
+        derivative_class = argparse.Action
+        if isinstance(action, argparse.Action):
+            derivative_class = action
+        elif isinstance(action, str):
+            derivative_class = self.command_loader.cli_ctx.invocation.parser._registries['action'][action]  # pylint: disable=protected-access
+
+        class DeprecatedArgumentAction(derivative_class):
+
+            def __init__(self, **kwargs):
+                super(DeprecatedArgumentAction, self).__init__(**kwargs)
+
+            def __call__(self, parser, namespace, values, option_string=None):
+                if not hasattr(namespace, '_argument_deprecations'):
+                    setattr(namespace, '_argument_deprecations', [deprecate_info])
+                else:
+                    namespace._argument_deprecations.append(deprecate_info)  # pylint: disable=protected-access
+                try:
+                    super(DeprecatedArgumentAction, self).__call__(parser, namespace, values, option_string)
+                except NotImplementedError:
+                    pass
+
+        return DeprecatedArgumentAction
+
+    def _get_opt_deprecation_action(self, kwargs, deprecated_options):
+
+        if not isinstance(deprecated_options, list):
+            deprecated_options = [deprecated_options]
+
+        action = kwargs.get('action', None)
+        derivative_class = argparse.Action
+        if isinstance(action, argparse.Action):
+            derivative_class = action
+        elif isinstance(action, str):
+            derivative_class = self.command_loader.cli_ctx.invocation.parser._registries['action'][action]  # pylint: disable=protected-access
+
+        class DeprecatedOptionAction(derivative_class):
+
+            def __init__(self, **kwargs):
+                super(DeprecatedOptionAction, self).__init__(**kwargs)
+
+            def __call__(self, parser, namespace, values, option_string=None):
+                deprecated_opt = next((x for x in deprecated_options if option_string == x.target), None)
+                if deprecated_opt:
+                    if not hasattr(namespace, '_argument_deprecations'):
+                        setattr(namespace, '_argument_deprecations', [deprecated_opt])
+                    else:
+                        namespace._argument_deprecations.append(deprecated_opt)  # pylint: disable=protected-access
+                try:
+                    super(DeprecatedOptionAction, self).__call__(parser, namespace, values, option_string)
+                except NotImplementedError:
+                    pass
+
+        return DeprecatedOptionAction
+
+    def deprecate(self, **kwargs):
+
+        def _get_deprecated_arg_message(self):
+            message = "{} '{}' has been deprecated and will be removed ".format(
+                self.object_type, self.target).capitalize()
+            if self.expiration:
+                message += "in version '{}'. ".format(self.expiration)
+            else:
+                message += 'in a future release. '
+            message += "Use '{}' instead. ".format(self.redirect) if self.redirect else ''
+            return message.lstrip()
+
+        target = kwargs.get('target', '')
+        kwargs['object_type'] = 'option' if target.startswith('-') else 'argument'
+        kwargs['message_func'] = _get_deprecated_arg_message
+        return Deprecated(self.command_loader.cli_ctx, **kwargs)
+
     def argument(self, argument_dest, arg_type=None, **kwargs):
         """ Register an argument for the given command scope using a knack.arguments.CLIArgumentType
 
@@ -153,19 +228,26 @@ class ArgumentsContext(object):
         :param kwargs: Possible values: `options_list`, `validator`, `completer`, `nargs`, `action`, `const`, `default`,
                        `type`, `choices`, `required`, `help`, `metavar`. See /docs/arguments.md.
         """
+        deprecate_info = kwargs.get('deprecate_info', None)
+        if deprecate_info:
+            deprecate_info.target = deprecate_info.target or argument_dest
+            kwargs['action'] = self._get_arg_deprecation_action(kwargs, deprecate_info)
+        deprecated_opts = [x for x in kwargs.get('options_list', []) if isinstance(x, Deprecated)]
+        if deprecated_opts:
+            kwargs['action'] = self._get_opt_deprecation_action(kwargs, deprecated_opts)
         self.command_loader.argument_registry.register_cli_argument(self.command_scope,
                                                                     argument_dest,
                                                                     arg_type,
                                                                     **kwargs)
 
-    def ignore(self, argument_dest):
+    def ignore(self, argument_dest, **kwargs):
         """ Register an argument with type knack.arguments.ignore_type (hidden/ignored)
 
         :param argument_dest: The destination argument to apply the ignore type to
         :type argument_dest: str
         """
         dest_option = ['--__{}'.format(argument_dest.upper())]
-        self.argument(argument_dest, arg_type=ignore_type, options_list=dest_option)
+        self.argument(argument_dest, arg_type=ignore_type, options_list=dest_option, **kwargs)
 
     def extra(self, argument_dest, **kwargs):
         """Register extra parameters for the given command. Typically used to augment auto-command built
@@ -176,6 +258,13 @@ class ArgumentsContext(object):
         :param kwargs: Possible values: `options_list`, `validator`, `completer`, `nargs`, `action`, `const`, `default`,
                        `type`, `choices`, `required`, `help`, `metavar`. See /docs/arguments.md.
         """
+        deprecate_info = kwargs.get('deprecate_info', None)
+        if deprecate_info:
+            deprecate_info.target = deprecate_info.target or argument_dest
+            kwargs['action'] = self._get_arg_deprecation_action(kwargs, deprecate_info)
+        deprecated_opts = [x for x in kwargs.get('options_list', []) if isinstance(x, Deprecated)]
+        if deprecated_opts:
+            kwargs['action'] = self._get_opt_deprecation_action(kwargs, deprecated_opts)
         self.command_loader.extra_argument_registry[self.command_scope][argument_dest] = CLICommandArgument(
             argument_dest, **kwargs)
 
--- a/knack/arguments.py
+++ b/knack/arguments.py
@@ -143,75 +143,80 @@ class ArgumentsContext(object):
     def __exit__(self, exc_type, exc_val, exc_tb):
         pass
 
-    def _get_arg_deprecation_action(self, kwargs, deprecate_info):
-
+    def _get_parent_class(self, **kwargs):
         # wrap any existing action
         action = kwargs.get('action', None)
-        derivative_class = argparse.Action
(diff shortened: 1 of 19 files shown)


derekbekoe:
@ mention me when CI passes and I can take a look!

tjprescott:
@derekbekoe finally passing! 😓

tjprescott:
@williexu ping

knack/deprecation.py
```diff
@@ -0,0 +1,179 @@
...
+        self.expiration = expiration
+
+        def _default_get_message(self):
+            message = "This {} has been deprecated and will be removed ".format(self.object_type)
+            if self.expiration:
+                message += "in version '{}'. ".format(self.expiration)
+            else:
+                message += 'in a future release. '
+            message += "Use '{}' instead. ".format(self.redirect) if self.redirect else ''
+            return message.lstrip()
```
williexu: A majority of this is replicated elsewhere

knack/deprecation.py
```diff
@@ -0,0 +1,179 @@
...
+
+        # pylint: disable=unused-argument
+        def _default_get_tag(self):
+            return DEFAULT_DEPRECATED_TAG
+
+        self._get_tag = tag_func or _default_get_tag
```
williexu: Why is `_default_get_tag()` needed? Why not `tag_func or DEFAULT_DEPRECATED_TAG`

knack/invocation.py
```diff
@@ -139,6 +145,28 @@ def execute(self, args):
...
+
+        if implicit_deprecate_info:
+            deprecate_kwargs = implicit_deprecate_info.__dict__.copy()
+            deprecate_kwargs['object_type'] = 'command'
+            del deprecate_kwargs['_get_tag']
+            del deprecate_kwargs['_get_message']
+            deprecations.append(ImplicitDeprecated(**deprecate_kwargs))
+
+        for d in deprecations:
+            logger.warning(str(d.message))
```
williexu: Is the `str()` necessary

knack/parser.py
```diff
@@ -112,20 +141,22 @@ def load_command_table(self, cmd_tbl):
...
                 _command_validator=command_validator,
                 _argument_validators=argument_validators,
                 _parser=command_parser)
 
-    def _get_subparser(self, path):
+    def _get_subparser(self, path, group_table=None):
         """For each part of the path, walk down the tree of
         subparsers, creating new ones if one doesn't already exist.
         """
+        group_table = group_table or {}
```
williexu: `def _get_subparser(self, path, group_table={}):` would remove the need for this

knack/parser.py
```diff
@@ -112,20 +141,22 @@ def load_command_table(self, cmd_tbl):
...
         for length in range(0, len(path)):
-            parent_subparser = self.subparsers.get(tuple(path[0:length]), None)
+            parent_path = path[0:length]
```
williexu: path[:length]

knack/parser.py
```diff
@@ -135,13 +166,25 @@ def _get_subparser(self, path):
...
                 # subcmd2 and so on), we know we can always back up one step and
                 # add a subparser if one doesn't exist
-                grandparent_subparser = self.subparsers[tuple(path[0:length - 1])]
-                new_parser = grandparent_subparser.add_parser(path[length - 1], cli_help=self.cli_help)
+                command_group = group_table.get(' '.join(parent_path))
+                if command_group:
+                    deprecate_info = command_group.group_kwargs.get('deprecate_info', None)
+                    if deprecate_info and deprecate_info.expired():
+                        continue
+                grandparent_path = path[0:length - 1]
```
williexu: path[:length - 1]

knack/arguments.py
```diff
@@ -143,6 +143,81 @@ def __enter__(self):
...
+            def __call__(self, parser, namespace, values, option_string=None):
+                deprecated_opt = next((x for x in deprecated_options if option_string == x.target), None)
+                if deprecated_opt:
+                    if not hasattr(namespace, '_argument_deprecations'):
+                        setattr(namespace, '_argument_deprecations', [deprecated_opt])
+                    else:
+                        namespace._argument_deprecations.append(deprecated_opt)  # pylint: disable=protected-access
+                try:
+                    super(DeprecatedOptionAction, self).__call__(parser, namespace, values, option_string)
+                except NotImplementedError:
```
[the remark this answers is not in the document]
tjprescott: If the superclass wasn't callable.

knack/invocation.py
tjprescott: Yes.

knack/parser.py
tjprescott: This is a big no-no in Python. 
http://effbot.org/zone/default-values.htm

knack/deprecation.py
tjprescott: Because DEFAULT_DEPRECATED_TAG is a string and thus not callable.  `self._get_tag` needs to be callable. I could use a lambda, but figured this would be a little clearer.

knack/deprecation.py
tjprescott: Good point. I'll see if I can abstract this out.

knack/invocation.py
williexu: Doesn't the logger.warning method call the string method of whatever is passed?
If it doesn't it probably should..

knack/deprecation.py
williexu: If it needs to be callable, I think a lambda would be better for something this simple.
That being said, why does this need to be callable?

knack/deprecation.py
tjprescott: Reviewing these cases, I'm not sure there's much value in templatizing this part.

knack/invocation.py
tjprescott: It doesn't and it's not under our control--hence the call to `str(...)`

knack/deprecation.py
tjprescott: Considering I used a lambda in another case, there's no harm using it here.  Looking at the current use cases, it could simply be a string, but since this is a framework, making it callable allows people to provide custom logic for future cases.

tjprescott:
@williexu all the tests are in place (they pass when run individually so there's some fun parallel nonsense going on) and I've made your changes. I'll fix up the test issue (related to helps...) but please take another look.

knack/deprecation.py
```diff
@@ -0,0 +1,180 @@
...
+        v2_comps = v2.split('.')
+        if len(v1_comps) != len(v2_comps):
+            from knack.util import CLIError
+            raise CLIError("Unable to compare version '{}' to version '{}'. Check version format.".format(v1, v2))
+        for i, _ in enumerate(v1_comps):
+            if v1_comps[i] < v2_comps[i]:
+                return True
+            elif v1_comps[i] > v2_comps[i]:
+                return False
+        return False
```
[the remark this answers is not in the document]
tjprescott: I believe it is on the hot path, but I'll try it out. I didn't realize it was a thing.

knack/help.py
```diff
@@ -253,155 +332,252 @@ def __init__(self, _data):
...
+        choice_str = u'  Allowed values: {0}.'.format(', '.join(sorted([str(x) for x in p.choices]))) \
             if p.choices else ''
-        default_str = '  Default: {0}.'.format(p.default) \
+        default_str = u'  Default: {0}.'.format(p.default) \
             if p.default and p.default != argparse.SUPPRESS else ''
-        value_sources_str = '  Values from: {0}.'.format(', '.join(p.value_sources)) \
+        value_sources_str = u'  Values from: {0}.'.format(', '.join(p.value_sources)) \
             if p.value_sources else ''
-        return '{0}{1}{2}'.format(choice_str, default_str, value_sources_str)
+        return u'{0}{1}{2}'.format(choice_str, default_str, value_sources_str)
```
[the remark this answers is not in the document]
tjprescott: Added due to Python 2/3 frustrations.  I might be able to take it back out.

knack/help.py
```diff
@@ -410,13 +586,23 @@ def __init__(self, cli_ctx=None, privacy_statement='', welcome_message=''):
         :type privacy_statement: str
         :param welcome_message: A welcome message for the CLI
         :type welcome_message: str
+        :param group_help_cls: Class to use for formatting group help.
+        :type group_help_cls: HelpFile
+        :param command_help_cls: Class to use for formatting command help.
+        :type command_help_cls: HelpFile
```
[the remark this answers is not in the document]
tjprescott: No, because you can specify your own class (which won't necessarily be of those types) but it should derive from HelpFile.

knack/deprecation.py
```diff
@@ -0,0 +1,180 @@
...
+        """
+        self.cli_ctx = cli_ctx
+        self.object_type = object_type
+        self.target = target
+        self.redirect = redirect
+        self.hide = hide
+        self.expiration = expiration
+
+        def _default_get_message(self):
+            line1 = "This {} has been deprecated and will be removed ".format(self.object_type)
```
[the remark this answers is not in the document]
tjprescott: They are pretty similar, but I don't see a good way to templatize them that would be used in other scenarios.

knack/commands.py
```diff
@@ -259,9 +266,16 @@ def command(self, name, handler_name, **kwargs):
                        `formatter_class`, `table_transformer`, `deprecate_info`, `validator`, `confirmation`.
         """
         command_name = '{} {}'.format(self.group_name, name) if self.group_name else name
-        command_kwargs = copy.deepcopy(self.group_kwargs)
+        command_kwargs = self.group_kwargs.copy()
```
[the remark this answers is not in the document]
tjprescott: I was able implement deepcopy for the Deprecated class. I do a shallow copy of the CLI context though, which seems appropriate.

knack/parser.py
```diff
@@ -93,11 +116,17 @@ def load_command_table(self, cmd_tbl):
                                                   help_file=metadata.help,
                                                   formatter_class=fc,
                                                   cli_help=self.cli_help)
-
+            command_parser.cli_ctx = self.cli_ctx
             command_validator = metadata.validator
             argument_validators = []
             argument_groups = {}
-            for arg in metadata.arguments.values():
+            for _, arg in metadata.arguments.items():
```
williexu: Why iterate over items() if you don't need to?

knack/help.py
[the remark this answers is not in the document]
tjprescott: Actually it has to do with fixing an issue in CLI where displaying help fails when the text contains a non-ASCII character.

tjprescott:
@derekbekoe I fixed the issue with deepcopy and colorama imports. Also used parse_version instead of a custom version check.  Please take a second look... oh, and you too @williexu!!! 👿

knack/parser.py
```diff
@@ -40,7 +41,23 @@ def _add_argument(obj, arg):
         """ Only pass valid argparse kwargs to argparse.ArgumentParser.add_argument """
         options_list = arg.options_list
         argparse_options = {name: value for name, value in arg.options.items() if name in ARGPARSE_SUPPORTED_KWARGS}
-        return obj.add_argument(*options_list, **argparse_options)
+        scrubbed_options_list = []
+        for item in options_list:
```
williexu: `for item in arg.options_list:`
remove line 42

knack/invocation.py
```diff
@@ -139,6 +147,29 @@ def execute(self, args):
...
+        if implicit_deprecate_info:
+            deprecate_kwargs = implicit_deprecate_info.__dict__.copy()
+            deprecate_kwargs['object_type'] = 'command'
+            del deprecate_kwargs['_get_tag']
+            del deprecate_kwargs['_get_message']
+            deprecations.append(ImplicitDeprecated(**deprecate_kwargs))
+
+        for d in deprecations:
+            msg = str(d.message)
+            print(msg, file=sys.stderr)
```
williexu: `print()` calls `str()` on its arguments

knack/deprecation.py
```diff
@@ -0,0 +1,188 @@
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+from six import string_types as STRING_TYPES
+
+from knack.log import get_logger
+
+logger = get_logger(__name__)
```
williexu: don't believe you use the logger anywhere

knack/invocation.py
```diff
@@ -17,6 +21,9 @@
 from .help import CLIHelp
 
 
+logger = get_logger(__name__)
```
williexu: Don't think you use this?

knack/invocation.py
```diff
@@ -139,6 +147,29 @@ def execute(self, args):
 
         params = self._filter_params(parsed_args)
 
+        cmd = parsed_args.func
+        deprecations = [] + getattr(parsed_args, '_argument_deprecations', [])
```
williexu: you're already giving an empty list as the default for `getattr()`

knack/deprecation.py
```diff
@@ -0,0 +1,188 @@
...
+        def _default_get_message(self):
+            line1 = "This {} has been deprecated and will be removed ".format(self.object_type)
+            if self.expiration:
+                line1 += "in version '{}'.".format(self.expiration)
+            else:
+                line1 += 'in a future release.'
+            lines = [line1]
+            if self.redirect:
+                lines.append("Use '{}' instead.".format(self.redirect))
+            return ' '.join(lines)
```
williexu: If lines is only going to have 1 or 2 elements, why not just add to the string if self.redirect?

knack/arguments.py
```diff
@@ -143,6 +143,89 @@ def __enter__(self):
...
+            line1 = "{} '{}' has been deprecated and will be removed ".format(
+                self.object_type, self.target).capitalize()
+            if self.expiration:
+                line1 += "in version '{}'.".format(self.expiration)
+            else:
+                line1 += 'in a future release.'
+            lines = [line1]
+            if self.redirect:
+                lines.append("Use '{}' instead.".format(self.redirect))
+            return ' '.join(lines)
```
williexu: why not just add to line1 if `self.redirect`?
lines will only have 1 or 2 elements.

Or, create the list at the beginning of this method, seems weird that you interchange between adding and then creating the list after the second line is added.

tjprescott:
@derekebekoe we can't really delay the import of parse version because you have to parse the versions to know whether the item in question should be added to the parser. Expired items are not added to the parser.

knack/invocation.py
tjprescott: Nope. It did use it, which was the source of all the horribleness. So happy to get rid of it.

knack/deprecation.py
tjprescott: TL/DR -> Trust me 😆 

Because you have to have annoying logic to include a space or not, and you end up with a whitespace nightmare. This ensures there is always and only ONE space between lines. I tried other schemes and it was a constant headache getting unit tests to pass due to "off-by-one" errors related exactly to trying to just add to strings.

williexu: okay I trust you

tjprescott: Actually, I did you suggestion and it worked fine.  Having unit tests is a good thing 👍

derekbekoe:
@tjprescott That would mean that we'd have to revisit using parse_version then or find another way to determine an older version?
Because we can't have a 0.5 second overhead in knack.

derekbekoe:
Let me know if you want to discuss offline.

knack/invocation.py
```diff
@@ -139,6 +145,30 @@ def execute(self, args):
...
+            deprecate_kwargs = implicit_deprecate_info.__dict__.copy()
+            deprecate_kwargs['object_type'] = 'command'
+            del deprecate_kwargs['_get_tag']
+            del deprecate_kwargs['_get_message']
+            deprecations.append(ImplicitDeprecated(**deprecate_kwargs))
+
+        colorama.init()
+        for d in deprecations:
+            print(d.message, file=sys.stderr)
+            colorama.deinit()
```
williexu: `colorama.deinit()` should be put after the for loop. You aren't reiniting colorama for each deperecation.

tjprescott: Good catch. Thanks!

PR #88, opened 2018-07-26
tab output: disable number parse

yugangw-msft:
See https://github.com/Azure/azure-cli/issues/6878
By default, we should not let tabulate auto-parse number inside a string which would become confusing for being unpredictable. If we receive user voices for auto-parse(very unlikely), we can find a way for applications to opt in.


yugangw-msft:
@troydai , can you take a look?

troydai:
Fixes CI before merge.

PR #96, opened 2018-09-13
Add option for setting config file name

Christina-Kang:
Add ability to override _CONFIG_FILE_NAME

Adding this because I wanted to use this code to write to a file other than config, in the same format, since the data I wanted to write had more to do with state than with configuration.


Christina-Kang:
@tjprescott requesting a review on this PR. Thank you! :)

Christina-Kang:
Requesting a review. Thank you! @yugangw-msft @derekbekoe

PR #97, opened 2018-09-17
Add new command invoke_and_return_output

Christina-Kang:
Add command invoke_and_return_output which gives users an option to get back as a string the output from the invoke command. This gives an easy way to get back the returned result (or exception message and stacktrace) to the caller should it be needed, while leaving other behavior the same.

The previous behavior of the invoke command is not altered.


Christina-Kang:
Requesting a review. Thank you! @yugangw-msft @derekbekoe

Christina-Kang:
Adding @tjprescott to request a review list. Thank you!!

tjprescott:
I believe the change introduced in Knack 0.5.0 will let you accomplish your scenario, so I will close this.

PR #108, opened 2018-10-13
Add Python3.7 support

daneoshiga:
To add python3.7 to tox check I had to upgrade pylint.
Doing so brought new checks

Added R0205 and R1717 to disable list because they're related to features only available on python3.7

Applied R1705: Unnecessary "else" after "return" (no-else-return) as needed.


tjprescott:
@daneoshiga please resolve the conflict, and thank you for the contribution!

daneoshiga:
@tjprescott conflict resolved

PR #124, opened 2018-11-07
Add features from Azure CLI into Knack

tjprescott:


tjprescott:
@williexu I'll make a PR for the CLI to use this version and some of the logic will be removed (some will still need to be overridden).  Primarily though, these are features that partner teams have requested and/or features I want to use in azdev.

adewaleo:
This also looks good to me! (y)

PR #146, opened 2019-03-29
Command suggestion feature from CLI 2.0

### LINKED ISSUE #131: Lift command suggestion feature from CLI 2.0.
CLI 2.0 has a feature where a mistyped command will offer suggestions of what you might have meant, which would be nice to have in Knack.

tjprescott:
Closes #131. Add the feature from Azure CLI where the CLI will try to suggest the closest alternative when the user mistypes a command.


tjprescott:
@limingu yes, my PR to update the CLI to use this version of knack will remove the duplicate code.

tjprescott:
@limingu it turns out that the version in the CLI needs to remain because it is wired for telemetry and has additional logic based on command_source, which is not in knack. But now it means that knack-based CLIs will have the command suggestions.

PR #148, opened 2019-04-19
local context: support chained config file

limingu:


knack/config.py
```diff
@@ -47,36 +47,126 @@ def __init__(self, config_dir=None, config_env_var_prefix=None, config_file_name
...
+        config_dir_name = os.path.basename(self.config_dir)
+        while current_dir:
+            current_config_dir = os.path.join(current_dir, config_dir_name)
+            if (os.path.normcase(os.path.normpath(current_config_dir)) ==
+                    os.path.normcase(os.path.normpath(self.config_dir))):
+                break
+            if os.path.isdir(current_config_dir):
+                self.config_file_chain.append(_ConfigFile(current_config_dir,
+                                                          os.path.join(current_config_dir, configuration_file_name)))
+            if current_dir == os.path.dirname(current_dir):
```
yugangw-msft: I assume this means we are getting to the root drive? if yes, I suggest call out in the comment

knack/config.py
```diff
@@ -47,36 +47,126 @@ def __init__(self, config_dir=None, config_env_var_prefix=None, config_file_name
...
-            env = self.env_var_name(section, option)
-            return os.environ[env] if env in os.environ else self.config_parser.get(section, option)
-        except (configparser.NoSectionError, configparser.NoOptionError):
-            if fallback is _UNSET:
-                raise
-            return fallback
+            self.config_parser.add_section(section)
+        except configparser.DuplicateSectionError:
+            pass
+        self.config_parser.set(section, option, value)
```
yugangw-msft: do we still need this? I thought `_ConfigFile.set_value` should handle it, no?

knack/config.py
```diff
@@ -47,36 +47,126 @@ def __init__(self, config_dir=None, config_env_var_prefix=None, config_file_name
...
+        return float(self.get(section, option, fallback, use_local_config))
+
+    def getboolean(self, section, option, fallback=_UNSET, use_local_config=False):
+        val = str(self.get(section, option, fallback, use_local_config))
+        if val.lower() not in CLIConfig._BOOLEAN_STATES:
+            raise ValueError('Not a boolean: {}'.format(val))
+        return CLIConfig._BOOLEAN_STATES[val.lower()]
 
-    def get(self, section, option, fallback=_UNSET):
+    def set_value(self, section, option, value, use_local_config=False):
```
yugangw-msft: looks  like the `set` method is removed, is it on purpose?

knack/config.py
```diff
@@ -47,36 +47,126 @@ def __init__(self, config_dir=None, config_env_var_prefix=None, config_file_name
         self._env_var_format = '{}{}'.format(env_var_prefix, '{section}_{option}')
         self.defaults_section_name = CLIConfig._CONFIG_DEFAULTS_SECTION
         self.config_parser.read(self.config_path)
+        self.config_file_chain = []
```
tjprescott: I would recommend `self._config_file_chain`. I see this as an implementation detail and would avoid making it public.

knack/config.py
```diff
@@ -47,36 +47,126 @@ def __init__(self, config_dir=None, config_env_var_prefix=None, config_file_name
...
+                break
+            current_dir = os.path.dirname(current_dir)
+        self.config_file_chain.append(_ConfigFile(self.config_dir, self.config_path))
 
     def env_var_name(self, section, option):
         return self._env_var_format.format(section=section.upper(),
                                            option=option.upper())
 
-    def has_option(self, section, option):
+    def has_option(self, section, option, use_local_config=False):
```
tjprescott: It seems to me that whether the user is using local config should be a property of the CLIConfig object and set during `__init__` rather than specified on each and every method call. Is there some scenario where it would make sense to `get` a property from local config but set it NOT using the local context?  If that was the case, I would still suggest making `use_local_config=None` and then pick the default value from the CLIConfig's property.

knack/config.py
limingu: Added comments.

knack/config.py
limingu: Updated

knack/config.py
limingu: Updated.

knack/config.py
limingu: Reomved

knack/config.py
limingu: Yes, only set_value is needed.

PR #149, opened 2019-04-24
Bump the version for new release

limingu:


setup.py
```diff
@@ -9,7 +9,7 @@
 from codecs import open
 from setuptools import setup, find_packages
 
-VERSION = '0.5.4'
+VERSION = '0.6.0'
```
marstr: Sorry, I don't have a lot of context on the changes that are instigating this version bump. Are there breaking changes to knack?

limingu: Yes, there are breaking changes for the config. It removed config_parser and set method in the last local context changes

tjprescott: Why?

limingu: Because originally only one config_parser and one config file. Now there is a list of config files. The config parser is in each config file class. So no need the config_parser for the config class itself.

yugangw-msft: It was old design defects that I suggested to fix instead of keeping it hence there are breaking change. Any concerns?

yugangw-msft: the related commit is https://github.com/Microsoft/knack/pull/148/commits/3bf5f6f7882a782da83647a71a759917ba82b2a8

tjprescott: Not really I suppose.

PR #152, opened 2019-05-21
Command/Argument Preview Mechanism

tjprescott:
Allows command authors to designate that commands, command groups, and arguments are in a preview state, similar to the existing deprecation mechanism.

### DIFF
--- a/knack/invocation.py
+++ b/knack/invocation.py
@@ -10,6 +10,7 @@ import sys
 from collections import defaultdict
 
 from .deprecation import ImplicitDeprecated, resolve_deprecate_info
+from .preview import ImplicitPreviewItem, resolve_preview_info
 from .util import CLIError, CtxTypeError, CommandResultItem, todict
 from .parser import CLICommandParser
 from .commands import CLICommandsLoader
@@ -117,6 +118,7 @@ class CommandInvoker(object):
             err = sys.exc_info()[1]
             getattr(parsed_ns, '_parser', self.parser).validation_error(str(err))
 
+    # pylint: disable=too-many-statements
     def execute(self, args):
         """ Executes the command invocation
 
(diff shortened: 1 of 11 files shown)


marstr:
Just out of curiosity, how easy would it be to tack on a `hidden` option? I'd love to have `self-test` stop telling me that it was going to be deleted soon if it's easy.

knack/invocation.py
```diff
@@ -180,9 +184,13 @@ def execute(self, args):
             del deprecate_kwargs['_get_message']
             deprecations.append(ImplicitDeprecated(**deprecate_kwargs))
 
+        # TODO: search for implicit preview?
```
marstr: Please clarify a little, what may constitute an implicit preview?

knack/preview.py
```diff
@@ -0,0 +1,109 @@
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+from six import string_types as STRING_TYPES
+
+from .util import ColorizedString
+
+_PREVIEW_TAG = '[Preview]'
```
marstr: nit: my vote would be for lowercase on this, but it's a weak opinion.

knack/invocation.py
marstr: I see the class name below, so I'm guessing you mean differentiating behavior. However, I'm still not sure what would make something implicitly a preview. :)

tjprescott: Actually, I do need to pull this code in from the CLI. You do need to have "implicit preview" so that if DLS is in preview, all of the subgroups and commands under it also recognize they are in preview.

knack/preview.py
tjprescott: I wanted to keep it consistent with `[Deprecated]`

marstr: Love consistency, thanks for tell me. :)

knack/invocation.py
marstr: Got it, may be worth adding that to a doc comment somewhere. :)

tjprescott:
@marstr I cleaned up the code to let develops simply use `is_preview=True`.

tjprescott:
Not sure why TravisCI isn't linked to GitHub now, but the build passed so I will merge.

PR #154, opened 2019-05-22
Fix target logic for preview mechanism

tjprescott:
Ensures that the longest option is used as target, not the shortest!

### DIFF
--- a/knack/arguments.py
+++ b/knack/arguments.py
@@ -256,7 +256,7 @@ class ArgumentsContext(object):
             # convert argument dest
             target = '--{}'.format(argument_dest.replace('_', '-'))
         elif options_list:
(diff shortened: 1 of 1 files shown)


knack/arguments.py
```diff
@@ -256,7 +256,7 @@ def _get_preview_arg_message(self):
             # convert argument dest
             target = '--{}'.format(argument_dest.replace('_', '-'))
         elif options_list:
-            target = sorted(options_list, key=len)[0]
+            target = sorted(options_list, key=len)[-1]
```
limingu: What will happen if the options_list has 2  longest items with the same length?

tjprescott: It will use either of them, which is fine in my mind. I just want to make sure it doesn't use the short option.

PR #171, opened 2019-12-03
Allow disabling color

### LINKED ISSUE #170: Allow disabling color
Related to #169. Reported in https://github.com/Azure/azure-cli/issues/6080. Also caused side effect of https://github.com/Azure/azure-cli/issues/9903.

If `stdout` is redirected and CLI displays a colored error message, the color in the console can't revert back to the default which is caused by https://github.com/tartley/colorama/issues/200.

Without proper a fix from Colorama, Knack should have the ability to disable color. This can be done by extending `_should_enable_color` by supporting env var like `KNACK_ENABLE_COLOR` or `KNACK_DISABLE_COLOR`.

https://github.com/microsoft/knack/blob/44ee1799c891213c90916f186ab9aee7b9ff8344/knack/log.py#L57-L64

jiasli:
Fixes #170, #169. Also reported in https://github.com/Azure/azure-cli/issues/6080 and causes side effect of https://github.com/Azure/azure-cli/issues/9903. 

If `stdout` is redirected and CLI displays a colored output, the color in the console can't revert back to the default because Colorama only resets color in `stdout`, but not `stderr` (tracked at https://github.com/tartley/colorama/issues/200, https://github.com/tartley/colorama/issues/218). This is affecting pip as well: https://github.com/pypa/pip/issues/6354 and pip uses `--no-color` to disable coloring https://github.com/pypa/pip/pull/4739. Azure CLI will add this option too.

Without proper a fix from Colorama, Knack can now 

- Disable color via `{CLI}_CORE_NO_COLOR` environment variable. 
- Use config
    ```
    [core]
    no_color=True
    ```

The PR removes Colorma from

1. **output**: output.py which replaces `out_file` with a Colorma stream and generates colorized output like jsonc
2. **log**: log.py which replaces `self.stream` with a Colorma stream and generates colorized logs
3. **help**: help.py which initializes Colorama for help message
4. **deprecation/preview warning**: invocation.py which initializes Colorama for deprecation and preview warning messages and util.py which generates colorized warning for `StatusTag`

`self.stream.isatty()` is removed because Colorama has its own mechanism to detect tty and treat PyCharm as a tty: https://github.com/tartley/colorama/blob/405b982f39efeaf697d3e2a6b90f9b913b38e97d/colorama/ansitowin32.py#L45

To test, 
1. Clone Knack to the local machine
2. Pull from my branch
3. Activate the virtual env of Azure CLI and run 

```sh
pip install --editable <path_to_knack>

# disable color
# cmd
set {CLI}_CORE_NO_COLOR=1
# bash
export {CLI}_CORE_NO_COLOR=1
# pwsh
$Env:{CLI}_CORE_NO_COLOR=1

# enable color
set {CLI}_CORE_NO_COLOR=

# test output color
az vm list --output jsonc

# test log color
az vm list --debug

# test help color
az vm create -h

# test deprecation warning color
az vm image accept-terms
```


Juliehzl:
@jiasli Should we consider upgrading knack version for the change?

jiasli:
Yes. But let's test it first. We will bump the version when other PRs are merged.

yungezz:
the change will disable color output from az cli totally?

jiasli:
Yes, Colorama won't be initialized with `KNACK_NO_COLOR` set to 1.

jsntcy:
- So if the user need disable color, does he/she need call 'set KNACK_NO_COLOR=1'?
- If user 'set KNACK_NO_COLOR=1'. is the behavior compatible as before for correct scenarios?

jiasli:
The user can either call `set KNACK_NO_COLOR=1` or set environment variable for the user or the system.

If the user runs `set KNACK_NO_COLOR=1`, only color is disabled with no change in the behavior, except that warnings will be prefixed with the level as a substitution for coloring, like `ERROR` for red 🟥, `WARNING` for yellow 🟨, `INFO` for green 🟩 and `DEBUG` for blue 🟦.

PR #173, opened 2020-02-17
Support yaml, yamlc output

jiasli:
Original PR https://github.com/Azure/azure-cli/pull/12234

This PR:

1. Move `yaml` and `yamlc` output [from Azure CLI](https://github.com/Azure/azure-cli/blob/5edbd822ee860d7ba5d5182811b68233e0846ba8/src/azure-cli-core/azure/cli/core/_output.py#L19-L40) to knack to unify output formats into one place
2. Add Unicode support in yaml output, like `こんにちは`
3. If `stdout` is a not tty, override jsonc with json, yamlc with yaml to remove color


knack/output.py
```diff
@@ -142,6 +164,8 @@ def out(self, obj, formatter=None, out_file=None):  # pylint: disable=no-self-us
                   file=out_file, end='')
 
     def get_formatter(self, format_type):  # pylint: disable=no-self-use
+        if not sys.stdout.isatty() and format_type in ['jsonc', 'yamlc']:
+            format_type = format_type[:-1]
```
qianwens: >format_type = format_type[:-1] [](start = 12, length = 30)

it works when you using string slice method to convert jsonc to json, but the code seems not easy to read, and if there is new type need convert like cjson->json, this code is not reusable. Can you explicitly map jsonc to json in your logic? Besides, I notice you add jsonc,ymalc in the OutputProducer._FORMAT_DICT above, what do you return different formatter here?

jiasli: Thank you for the suggestions. Refactored as requested. Also added test for removing color.

jiasli:
> Just one question. Even if this PR is merged, this won't work until you remove the inherited logic in azure-cli-core. Right? I mean this `get_formatter` part.

yes

PR #174, opened 2020-02-17
Drop python 2 and 3.5

jiasli:
Drop python 2 and 3.5 as Azure CLI has dropped them: https://github.com/Azure/azure-cli/pull/11363


.flake8
```diff
@@ -4,6 +4,10 @@ max-complexity = 10
...
     E126,
     E501,
+    E741,
+    E722,
     F401,
     F811,
-    C901
+    C901,
+    W503,
+    W504
```
jiasli: New entries are added to mute newer versions of flake8.

requirements.txt
```diff
@@ -1,13 +1,12 @@
...
+flake8==3.7.9
+jmespath==0.9.4
+mock==4.0.1
+pylint==2.3.0
+Pygments==2.5.2
+PyYAML==5.3
+six==1.14.0
+tabulate==0.8.6
+vcrpy==4.0.2
+pytest==5.3.5
```
jiasli: Updated using `pip freeze --requirement requirements.txt`

setup.py
```diff
@@ -39,12 +39,11 @@
         'Intended Audience :: Developers',
         'Intended Audience :: System Administrators',
         'Programming Language :: Python',
-        'Programming Language :: Python :: 2',
-        'Programming Language :: Python :: 2.7',
         'Programming Language :: Python :: 3',
         'Programming Language :: Python :: 3.5',
```
jiasli: `3.5` tag is still persisted, but not tested in CI.

tests/test_config.py
```diff
@@ -186,13 +186,15 @@ def test_set_config_value_file_permissions(self):
...
-        self.assertFalse(bool(file_mode & stat.S_IXOTH))
+        # only S_IRUSR and S_IWUSR are supported on Windows: https://docs.python.org/3.8/library/os.html#os.chmod
+        if os.name != 'nt':
+            self.assertFalse(bool(file_mode & stat.S_IXUSR))
+            self.assertFalse(bool(file_mode & stat.S_IRGRP))
+            self.assertFalse(bool(file_mode & stat.S_IWGRP))
+            self.assertFalse(bool(file_mode & stat.S_IXGRP))
+            self.assertFalse(bool(file_mode & stat.S_IROTH))
+            self.assertFalse(bool(file_mode & stat.S_IWOTH))
+            self.assertFalse(bool(file_mode & stat.S_IXOTH))
```
jiasli: Windows doesn't have Linux-like user/group/other concepts for `stat`. `rwx` is always the same for all scopes.

haroldrandom: image

```
>>> import sys
>>> sys.version
'3.8.0 (tags/v3.8.0:fa919fd, Oct 14 2019, 19:37:50) [MSC v.1916 64 bit (AMD64)]'
```

As far as I understand from the doc, the limitation only takes effect on `os.chmod`, however, `os.stat` works for me. The doc didn't mention that `os.stat` attributes won't be populated on Windows. 

I agreed on file mode checking across different OS platform, but if file mode checking differentiate on Windows, should we add additional support?

jiasli: I also didn't find the doc for `os.stat`'s behavior on Windows.

`st_mode=33206` is the default for a Windows file, which is `‭1000000 110 110 110‬` in binary. You can see `rwx` is the same `110` for different scopes. However you change it, it keeps this pattern. For example, if you change the file to read-only via `os.chmod('{}', stat.S_IRUSR)`, `os.stat` is `33060` which is `‭1000000 100 100 100‬` in binary. 

Checking permission on Windows requires `win32security` according to https://stackoverflow.com/a/12168268/2199657. This will over complicate the test. Since we only save the config file under `~/.azure`, we can assume Windows behaves correctly unless the system is compromised, but of course this is out of the scope of the app itself.

jiasli: I believe you are running this command on Cygwin which has additional file management logic. Running `os.chmod` and `os.stat` on bare Windows will repro my observation.

haroldrandom: No, on CMD

PR #176, opened 2020-02-28
Stop using random temp folders in tests

jiasli:
Originally, each time the full test is run (`pytest` or `tox`), it will generate ~20 temporary folders `tmp*` under 

1. User's temp folder: `C:\Users\{}\AppData\Local\Temp`
2. Working directory

This PR removes the usage of random temp folders and unify all tests to use `C:\Users\{}\AppData\Local\Temp\knack_temp` as user's temp folder and cleans the temp folder under the working directory. 

Note this change will invalidate parallel tests, but this is fine given the number of tests is small and we are not using parallel test anyway.

### DIFF
--- a/tests/util.py
+++ b/tests/util.py
@@ -9,9 +9,14 @@ except ImportError:
     from unittest import mock
 import sys
 import tempfile
+import shutil
+import os
 from six import StringIO
 
 from knack.cli import CLI, CLICommandsLoader, CommandInvoker
+
+TEMP_FOLDER_NAME = "knack_temp"
+
 
 def redirect_io(func):
 
@@ -30,7 +35,7 @@ def redirect_io(func):
 class MockContext(CLI):
 
     def __init__(self):
-        super(MockContext, self).__init__(config_dir=tempfile.mkdtemp())
+        super(MockContext, self).__init__(config_dir=new_temp_folder())
         loader = CLICommandsLoader(cli_ctx=self)
         invocation = mock.MagicMock(spec=CommandInvoker)
         invocation.data = {}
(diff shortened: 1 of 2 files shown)


haroldrandom:
Why would it invalidate parallel tests

jiasli:
Making subfolders under `C:\Users\{}\AppData\Local\Temp\knack_temp\` will still generate trash. They are generated in `__init__`. There is no good way to clear them, since [`__del__`](https://docs.python.org/3/reference/datamodel.html#object.__del__) is not guaranteed to be called. Actually I am not very concerned about `C:\Users\{}\AppData\Local\Temp\`. 

The trash folders under the working directory is worse. If one process only clears it's own temp folder, when it is forcefully stopped, `tearDown` won't be called and that leaves some trash. However, if we let `tearDown`  clear all temp folders with a prefix, processes with interfere with each other.

No way to solve this. So let's keep it simple and single process.

tests/util.py
```diff
@@ -44,5 +49,13 @@ def get_cli_version(self):
         return '0.1.0'
 
     def __init__(self, **kwargs):
-        kwargs['config_dir'] = tempfile.mkdtemp()
+        kwargs['config_dir'] = new_temp_folder()
         super(DummyCLI, self).__init__(**kwargs)
+
+
+def new_temp_folder():
+    temp_dir = os.path.join(tempfile.gettempdir(), TEMP_FOLDER_NAME)
```
MyronFanQiu: Does it mean that user cannot run two tests simultaneously?

jiasli: No

jiasli:
> if it's indeed a temp folder, I think it's suitable to use tempfile. We just need to clean these folders as you did in this PR.

I explained it in the previous comment.

PR #180, opened 2020-03-11
Add experimental tag

jiasli:
The experimental tag (same as preview) appears in following locations:

### Help of parent command group: [Tag]

```sh
> python examples\exapp2 -h

Subgroups:
    exp [Experimental] : An experimental command group.
```

### Help of itself: message

```sh
> python examples\exapp2 exp -h

Group
    exapp2 exp : An experimental command group.
        This command group is experimental and not covered by customer support. Please use with
        discretion.
```

### Help of parent command group: [Tag]

```sh
> python examples\exapp2 abc -h

Commands:
    last [Experimental] : List the last several letters in the alphabet.
```

### Help of itself: message

```sh
> python examples\exapp2 abc last -h

Command
    exapp2 abc last : List the last several letters in the alphabet.
        This command is experimental and not covered by customer support. Please use with
        discretion.
```

### Invocation: message

```sh
> python examples\exapp2 abc last
This command is experimental and not covered by customer support. Please use with discretion.
```

### Help of parent command: [Tag] + message

```sh
> python examples\exapp2 ga range -h

Arguments
    --end [Experimental] : The higher bound.  Default: 5.
        Argument '--end' is experimental and not covered by customer support. Please use with
        discretion.
```

### Invocation with this argument: message

```sh
> python examples\exapp2 ga range --end 3
Argument '--end' is experimental and not covered by customer support. Please use with discretion.
```

PR #181, opened 2020-03-16
Merge dev to master

jiasli:
This PR includes changes from these PRs:

- Allow disabling color (#171)
- Support `--only-show-errors` (#179)
- Add experimental tag (#180)

### DIFF
--- a/knack/log.py
+++ b/knack/log.py
@@ -54,24 +54,10 @@ class _CustomStreamHandler(logging.StreamHandler):
 
         return cls.COLOR_MAP.get(level, None)
 
-    def _should_enable_color(self):
-        try:
-            # Color if tty stream available
-            if self.stream.isatty():
-                return True
-        except AttributeError:
-            pass
-        return False
-
-    def __init__(self, log_level_config, log_format):
-        import platform
-        import colorama
-
+    def __init__(self, log_level_config, log_format, enable_color):
         logging.StreamHandler.__init__(self)
         self.setLevel(log_level_config)
-        if platform.system() == 'Windows':
-            self.stream = colorama.AnsiToWin32(self.stream).stream
-        self.enable_color = self._should_enable_color()
+        self.enable_color = enable_color
         self.setFormatter(logging.Formatter(log_format[self.enable_color]))
 
     def format(self, record):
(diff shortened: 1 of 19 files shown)


knack/log.py
```diff
@@ -140,22 +130,31 @@ def configure(self, args):
...
-        # Use max verbose level if too much verbosity specified.
-        return min(verbose_level, len(self.console_log_configs) - 1)
+        # arguments have higher precedence than config
+        if CLILogging.ONLY_SHOW_ERRORS_FLAG in args:
+            if CLILogging.DEBUG_FLAG in args or CLILogging.VERBOSE_FLAG in args:
+                raise CLIError("--only-show-errors can't be used together with --debug or --verbose")
+            self.cli_ctx.only_show_errors = True
+            return 1
+        if CLILogging.DEBUG_FLAG in args:
+            self.cli_ctx.only_show_errors = False
```
qianwens: >self.cli_ctx.only_show_errors = False [](start = 12, length = 37)

It seems not necessary, I think by default it is false

jiasli: This line is to make sure the config is overridden by `--debug`. If the config is `only_show_errors=True`, `--debug` will revert it to `False`.

knack/log.py
```diff
@@ -140,22 +130,31 @@ def configure(self, args):
...
+            return 4
+        if CLILogging.VERBOSE_FLAG in args:
+            self.cli_ctx.only_show_errors = False
+            return 3
+        if self.cli_ctx.only_show_errors:
+            return 1
```
qianwens: is this code possible reached, I think it already return 1 in ( if CLILogging.ONLY_SHOW_ERRORS_FLAG in args )

knack/log.py
qianwens: I see you has check debug cannot be set while only_show_errors set

---
In reply to: [394914515](https://github.com/microsoft/knack/pull/181#discussion_r394914515) [](ancestors = 394914515)

knack/log.py
jiasli: This takes effect when no explicit `--verbose`, `--debug` or `--only-show-warnings` is given, but `only_show_errors` is specified in the config. You may refer to `tests/test_log.py` for a enumeration of possible scenarios.

knack/log.py
jiasli: The check at L136 is only to check the argument `--only-show-errors`. It doesn't raise error when the config has the `only_show_errors=True`. In this case, `self.cli_ctx.only_show_errors` is reverted by the `--debug` argument.

PR #183, opened 2020-03-20
Add back python 3.5 in CI

jiasli:
Revert part of https://github.com/microsoft/knack/pull/174 to add python 3.5 back, as there are still many people using Python 3.5: https://w3techs.com/technologies/details/pl-python/3

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 sudo: false
 language: python
 python:
+- '3.5'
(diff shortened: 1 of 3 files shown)


jiasli:
Mock has dropped python 3.5: https://mock.readthedocs.io/en/latest/

> version 3.0.5 is the last version supporting Python 3.5 and lower.

PR #189, opened 2020-04-03
Remove python < 3.4 dependency

scarabeusiv:
As the oldest supported is python 3.5 it does not make sense to pull in enum34.


yonzhan:
python version

setup.py
```diff
@@ -48,7 +48,4 @@
     ],
     packages=['knack', 'knack.testsdk'],
     install_requires=DEPENDENCIES,
```
jiasli, line 50:
```suggestion
    install_requires=DEPENDENCIES
```

PR #199, opened 2020-04-21
Refine the timing to raise EVENT_CLI_POST_EXECUTE event and catch SystemExit

haroldrandom:
1.  make application be aware of the SystemExit exception
2.  Move a final event EVENT_CLI_POST_EXECUTE to `finally` statement to let application be aware of the final existing event and be able to register callbacks to do something.


jiasli:
Let's submit the PR for Azure CLI as well and review them together.

haroldrandom:
> Let's submit the PR for Azure CLI as well and review them together.

Added

Juliehzl:
Do we still need the change to close logger before https://github.com/Azure/azure-cli/blob/dev/src/azure-cli-core/azure/cli/core/azlogging.py#L183?

PR #201, opened 2020-04-30
Expose several remove functions in ConfigFile & Support write notice in config file

arrownj:
Previously, there are only `get` and `set` function in `_ConfigFile`, we can't remove section/option from a config file. Thus we have to set value to empty string if we want to remove a entry. Add `remove_section` and `remove_option` function here for the case we need to remove these entries.

Also add a notice property for ConfigFile, if we pass a notice, it will be wrote to head to the file.

### DIFF
--- a/knack/config.py
+++ b/knack/config.py
@@ -3,17 +3,12 @@
 # Licensed under the MIT License. See License.txt in the project root for license information.
 # --------------------------------------------------------------------------------------------
 import os
-import sys
 import stat
-from six.moves import configparser
+import configparser
 
 from .util import ensure_dir
 
 _UNSET = object()
-
-
-def get_config_parser():
-    return configparser.ConfigParser() if sys.version_info.major == 3 else configparser.SafeConfigParser()
 
 
 class CLIConfig(object):
@@ -152,11 +147,23 @@ class _ConfigFile(object):
     _BOOLEAN_STATES = {'1': True, 'yes': True, 'true': True, 'on': True,
                        '0': False, 'no': False, 'false': False, 'off': False}
 
-    def __init__(self, config_dir, config_path, config_notice=None):
+    def __init__(self, config_dir, config_path, config_comment=None):
+        """ Manages configuration options available in the CLI
+
+        :param config_dir: The directory to store config file
+        :type config_dir: str
+        :param config_path: The path of config file
+        :type config_path: str
+        :param config_comment: The comment which will be written into head of config file
+        :type config_comment: str
+
+        When 'config_comment' is given, each line should start with # or ;. For details about INI file comment,
+        you can reference this link https://docs.python.org/3/library/configparser.html#supported-ini-file-structure
+        """
         self.config_dir = config_dir
         self.config_path = config_path
-        self.config_notice = config_notice
-        self.config_parser = get_config_parser()
+        self.config_comment = config_comment
+        self.config_parser = configparser.ConfigParser()
         if os.path.exists(config_path):
             self.config_parser.read(config_path)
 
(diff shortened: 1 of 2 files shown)


knack/config.py
```diff
@@ -192,3 +201,32 @@ def set_value(self, section, option, value):
             pass
         config.set(section, option, value)
         self.set(config)
+
+    def remove_option(self, section, option):
+        config = get_config_parser()
+        config.read(self.config_path)
+        ret = False
```
jiasli, line 208: Let's avoid using general name `ret`, but `existed` as `configparser.py:935` does.

knack/config.py
```diff
@@ -141,14 +141,21 @@ def set_value(self, section, option, value):
     def set_to_use_local_config(self, use_local_config):
         self.use_local_config = use_local_config
 
+    def remove_option(self, section, option):
+        for config in self._config_file_chain if self.use_local_config else self._config_file_chain[-1:]:
+            if config.remove_option(section, option):
+                return True
```
jiasli, line 147: Not sure if it is a good idea to fall back to the parent config if `option` is not found. The other way is to only remove the option in this local config and keep the behavior same as `set_value`. 

Switch `use_local_config` like [`ConfiguredDefaultSetter`](https://github.com/Azure/azure-cli/blob/34f9033b96e6ea31f82626a9f2b152c1a061300e/src/azure-cli/azure/cli/command_modules/configure/custom.py#L150) does, or provide a scope parameter in `remove_option` to override `self.use_local_config`.

Again, not sure. I am open to this.

arrownj: It's a little different with `set_value`. I think `remove` may need map to `get`. As we can list values of parent config, I think it's reasonable to remove them in the same file.

knack/config.py
arrownj: thanks, updated.

knack/config.py
```diff
@@ -141,14 +141,21 @@ def set_value(self, section, option, value):
...
 
 class _ConfigFile(object):
     _BOOLEAN_STATES = {'1': True, 'yes': True, 'true': True, 'on': True,
                        '0': False, 'no': False, 'false': False, 'off': False}
 
-    def __init__(self, config_dir, config_path):
+    def __init__(self, config_dir, config_path, config_notice=None):
         self.config_dir = config_dir
         self.config_path = config_path
+        self.config_notice = config_notice
```
jiasli, line 158: Let's call it `config_comment`, and also add some comments about its functionality, including some examples, to make sure `;` is added correctly. 

```suggestion
        self.config_comment = config_comment
```

knack/config.py
```diff
@@ -141,15 +136,34 @@ def set_value(self, section, option, value):
...
+
+        :param config_dir: The directory to store config file
+        :type config_dir: str
+        :param config_path: The path of config file
+        :type config_path: str
+        :param config_comment: The comment which will be written into head of config file
+        :type config_comment: str
+
+        When 'config_comment' is given, each line should start with # or ;. For details about INI file comment,
+        you can reference this link https://docs.python.org/3/library/configparser.html#supported-ini-file-structure
```
jiasli, line 161:
```suggestion
        refer to https://docs.python.org/3/library/configparser.html#supported-ini-file-structure
```

knack/config.py
```diff
@@ -141,15 +136,34 @@ def set_value(self, section, option, value):
...
+        return False
+
 
 class _ConfigFile(object):
     _BOOLEAN_STATES = {'1': True, 'yes': True, 'true': True, 'on': True,
                        '0': False, 'no': False, 'false': False, 'off': False}
 
-    def __init__(self, config_dir, config_path):
+    def __init__(self, config_dir, config_path, config_comment=None):
+        """ Manages configuration options available in the CLI
```
jiasli, line 151:
```suggestion
        """ Manage configuration options available in the CLI
```

knack/config.py
```diff
@@ -141,15 +136,34 @@ def set_value(self, section, option, value):
...
+
+        :param config_dir: The directory to store config file
```
jiasli, line 153:
```suggestion
        :param config_dir: The directory to store the config file
```

knack/config.py
```diff
@@ -141,15 +136,34 @@ def set_value(self, section, option, value):
...
+        :type config_dir: str
+        :param config_path: The path of config file
```
jiasli, line 155:
```suggestion
        :param config_path: The path of the config file
```

knack/config.py
```diff
@@ -141,15 +136,34 @@ def set_value(self, section, option, value):
...
+        :type config_path: str
+        :param config_comment: The comment which will be written into head of config file
```
jiasli, line 157:
```suggestion
        :param config_comment: The comment which will be written into the head of the config file
```

PR #202, opened 2020-04-30
Add a default_value_source property in HelpParameter

arrownj:
Previously, both default value from function default and `az configure` will all display as `Default:`, we want to divide these different source. Add this property to save the default value source. It will be displayed in the help message.

### DIFF
--- a/knack/commands.py
+++ b/knack/commands.py
@@ -99,6 +99,7 @@ class CLICommand(object):  # pylint:disable=too-many-instance-attributes
             logger.info("Configured default '%s' for arg %s", config_value, arg.name)
             overrides.settings['default'] = DefaultStr(config_value)
             overrides.settings['required'] = False
+            overrides.settings['default_from'] = 'Configure'
 
     def load_arguments(self):
         if self.arguments_loader:
--- a/knack/commands.py
+++ b/knack/commands.py
@@ -99,7 +99,7 @@ class CLICommand(object):  # pylint:disable=too-many-instance-attributes
             logger.info("Configured default '%s' for arg %s", config_value, arg.name)
             overrides.settings['default'] = DefaultStr(config_value)
             overrides.settings['required'] = False
-            overrides.settings['default_from'] = 'Configure'
+            overrides.settings['default_value_source'] = 'Configure'
 
     def load_arguments(self):
         if self.arguments_loader:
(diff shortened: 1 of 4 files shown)


knack/commands.py
```diff
@@ -99,6 +99,7 @@ def _resolve_default_value_from_config_file(self, arg, overrides):
             logger.info("Configured default '%s' for arg %s", config_value, arg.name)
             overrides.settings['default'] = DefaultStr(config_value)
             overrides.settings['required'] = False
+            overrides.settings['default_from'] = 'Configure'
```
haroldrandom, line 102: What other else sources could be ?

arrownj: For local context, it's `Local Context`. If no value for this property, it will be treated as `Default`.

haroldrandom: What other values could be?

knack/parser.py
```diff
@@ -174,6 +175,7 @@ def load_command_table(self, command_loader):
                 param.deprecate_info = arg.deprecate_info
                 param.preview_info = arg.preview_info
                 param.experimental_info = arg.experimental_info
+                param.default_from = arg.default_from
```
jiasli, line 178: I prefer calling it `default_value_source` for better understanding.

knack/commands.py
arrownj: Currently, there are only 3 sources: Default, Configure, Local Context.

knack/parser.py
arrownj: thanks, updated.

knack/commands.py
```diff
@@ -99,6 +99,7 @@ def _resolve_default_value_from_config_file(self, arg, overrides):
             logger.info("Configured default '%s' for arg %s", config_value, arg.name)
             overrides.settings['default'] = DefaultStr(config_value)
             overrides.settings['required'] = False
+            overrides.settings['default_value_source'] = 'Configure'
```
jiasli, line 102: `Configure` is a verb. Maybe `config`?

```suggestion
            overrides.settings['default_value_source'] = 'config'
```

arrownj: This value will be shown in help message directly. The other two sources are `Default` and `Local Context`, so maybe `Config` is more reasonable ?

arrownj: Update to `Config`.

PR #220, opened 2020-09-28
[Config] Make config items case-insensitive

evelyn-ys:
## Description
Fix https://github.com/Azure/azure-cli/issues/14397: az configure --list-defaults should not be case sensitive for environment variables

## Testing Guide
preparation(need Azure CLI in advance)
```sh
az configure --defaults location=eastus
export AZURE_DEFAULTS_LOCATION=eastasia
```
test
```sh
az configure --list-defaults
```

results before
```
[
  {
    "name": "LOCATION",
    "source": "AZURE_DEFAULTS_LOCATION",
    "value": "eastasia"
  },
  {
    "name": "location",
    "source": "C:\\Users\\yishiwang\\.azure\\config",
    "value": "eastus"
  }
]
```

results after
```
[
  {
    "name": "LOCATION",
    "source": "AZURE_DEFAULTS_LOCATION",
    "value": "eastasia"
  }
]
```

### DIFF
--- a/knack/config.py
+++ b/knack/config.py
@@ -111,12 +111,12 @@ class CLIConfig(object):
         import re
         pattern = self.env_var_name(section, '.+')
         candidates = [(k.split('_')[-1], os.environ[k], k) for k in os.environ if re.match(pattern, k)]
-        result = {c[0]: c for c in candidates}
+        result = {c[0].lower(): c for c in candidates}
         for config in self._config_file_chain if self.use_local_config else self._config_file_chain[-1:]:
             try:
                 entries = config.items(section)
                 for name, value in entries:
-                    if name not in result:
+                    if name.lower() not in result:
                         result[name] = (name, value, config.config_path)
             except (configparser.NoSectionError, configparser.NoOptionError):
                 pass
--- a/knack/config.py
+++ b/knack/config.py
@@ -109,14 +109,14 @@ class CLIConfig(object):
 
     def items(self, section):
         import re
-        pattern = self.env_var_name(section, '.+')
+        pattern = self.env_var_name(section, '[0-9A-Z]+')
         candidates = [(k.split('_')[-1], os.environ[k], k) for k in os.environ if re.match(pattern, k)]
         result = {c[0].lower(): c for c in candidates}
         for config in self._config_file_chain if self.use_local_config else self._config_file_chain[-1:]:
             try:
                 entries = config.items(section)
                 for name, value in entries:
-                    if name.lower() not in result:
+                    if name not in result:
                         result[name] = (name, value, config.config_path)
             except (configparser.NoSectionError, configparser.NoOptionError):
                 pass
--- a/knack/config.py
+++ b/knack/config.py
@@ -109,8 +109,8 @@ class CLIConfig(object):
 
     def items(self, section):
         import re
-        pattern = self.env_var_name(section, '[0-9A-Z]+')
-        candidates = [(k.split('_')[-1], os.environ[k], k) for k in os.environ if re.match(pattern, k)]
+        pattern = self.env_var_name(section, '[0-9A-Z][0-9A-Z_]*')
+        candidates = [(k.replace(self.env_var_name(section, ''), ''), os.environ[k], k) for k in os.environ if re.match(pattern, k)]
         result = {c[0].lower(): c for c in candidates}
         for config in self._config_file_chain if self.use_local_config else self._config_file_chain[-1:]:
             try:
--- a/knack/config.py
+++ b/knack/config.py
@@ -109,9 +109,18 @@ class CLIConfig(object):
 
     def items(self, section):
         import re
-        pattern = self.env_var_name(section, '[0-9A-Z][0-9A-Z_]*')
-        candidates = [(k.replace(self.env_var_name(section, ''), ''), os.environ[k], k) for k in os.environ if re.match(pattern, k)]
-        result = {c[0].lower(): c for c in candidates}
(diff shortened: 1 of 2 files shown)


tests/test_config.py
```diff
@@ -169,6 +169,20 @@ def test_getboolean_error(self):
...
             self.cli_config.getboolean(section, option)
 
+    def test_items_case_insensitive(self):
+        section = 'MySection'
+        self.cli_config.set_value(section, 'option', 'value')
+        self.cli_config.set_to_use_local_config(True)
+        self.cli_config.set_value(section, 'Option', 'localValue')
+        items_result = self.cli_config.items(section)
+        self.assertEqual(len(items_result), 1)
+        self.assertEqual(items_result[0]['value'], 'localValue')
```
jiasli, line 179: Please add some comments for what is under testing.

tests/test_config.py
```diff
@@ -169,6 +169,20 @@ def test_getboolean_error(self):
...
+
+        with mock.patch.dict('os.environ', {self.cli_config.env_var_name(section, 'OPTION'): 'envValue'}):
```
jiasli, line 181: Better to use raw string, instead of `env_var_name` to test the functionality more thoroughly. Consider:

```py
self.assertEqual(env_var_name(), env_var_name())
```

knack/config.py
```diff
@@ -111,12 +111,12 @@ def items(self, section):
...
         pattern = self.env_var_name(section, '.+')
         candidates = [(k.split('_')[-1], os.environ[k], k) for k in os.environ if re.match(pattern, k)]
-        result = {c[0]: c for c in candidates}
+        result = {c[0].lower(): c for c in candidates}
         for config in self._config_file_chain if self.use_local_config else self._config_file_chain[-1:]:
             try:
                 entries = config.items(section)
                 for name, value in entries:
-                    if name not in result:
+                    if name.lower() not in result:
```
jiasli, line 119: We should force lower case for all configs in files, otherwise we need to change `get` and `set` as well.

knack/config.py
```diff
@@ -111,12 +111,12 @@ def items(self, section):
         import re
         pattern = self.env_var_name(section, '.+')
```
jiasli, line 112: Let's make `A-Z0-9` the only allows chars, so that we adhere to the doc https://docs.microsoft.com/en-us/cli/azure/azure-cli-configuration?view=azure-cli-latest#cli-configuration-values-and-environment-variables:

> Their corresponding environment variables are set as `AZURE_{section}_{name}`, in all caps.

knack/config.py
evelyn-ys: Shall we add force rule in `set_value`?

knack/config.py
```diff
@@ -109,9 +109,9 @@ def sections(self):
 
     def items(self, section):
         import re
-        pattern = self.env_var_name(section, '.+')
+        pattern = self.env_var_name(section, '[0-9A-Z]+')
```
jiasli, line 112: The option name can consist of multiple parts, like `AZURE_BATCHAI_STORAGE_ACCOUNT`.

knack/config.py
```diff
@@ -109,9 +109,9 @@ def sections(self):
 
     def items(self, section):
         import re
-        pattern = self.env_var_name(section, '.+')
-        candidates = [(k.split('_')[-1], os.environ[k], k) for k in os.environ if re.match(pattern, k)]
-        result = {c[0]: c for c in candidates}
+        pattern = self.env_var_name(section, '[0-9A-Z][0-9A-Z_]*')
+        candidates = [(k.replace(self.env_var_name(section, ''), ''), os.environ[k], k) for k in os.environ if re.match(pattern, k)]
```
jiasli, line 113: `replace` can cause unwanted effect if `self.env_var_name(section, '')` somehow appears in the middle of the env var. [`removeprefix`](https://docs.python.org/3/library/stdtypes.html#str.removeprefix) is a better choice but it is only available since Python 3.9.

Better to use [grouping](https://docs.python.org/3/howto/regex.html#grouping) of regex.

tests/test_config.py
```diff
@@ -169,6 +169,22 @@ def test_getboolean_error(self):
         with self.assertRaises(ValueError):
             self.cli_config.getboolean(section, option)
 
+    def test_items_case_insensitive(self):
+        section = 'MYSECTION'
+        envOption = '{}_{}_{}'.format(CLIConfig._DEFAULT_CONFIG_ENV_VAR_PREFIX, section, 'TEST_OPTION')
+        envVaule = 'envValue'
+        configOption = 'test_option'
+        configValue = 'value'
```
jiasli, line 177: The python convention is to use snake case.

knack/config.py
```diff
@@ -109,9 +109,9 @@ def sections(self):
 
     def items(self, section):
         import re
-        pattern = self.env_var_name(section, '.+')
-        candidates = [(k.split('_')[-1], os.environ[k], k) for k in os.environ if re.match(pattern, k)]
-        result = {c[0]: c for c in candidates}
+        pattern = self.env_var_name(section, '[0-9A-Z][0-9A-Z_]*')
```
jiasli, line 112: This will match `CLI_MYSECTION_T` part in `CLI_MYSECTION_Test_Option`, defeating our purpose of ignoring invalid env options like `CLI_MYSECTION_Test_Option`.

qwordy (approved):

tests/test_config.py
qwordy: > 
> 
> The python convention is to use snake case.

Agree

arrownj (approved):

evelyn-ys:
looks good now

jiasli (approved):

PR #223, opened 2020-10-30
[Tag] Allow customizing tag message

jiasli:
A singleton dict `status_tag_messages` is used to hold configurable `StatusTag` messages. Developers can override the content of `status_tag_messages` to customize the help and warning messages.

### DIFF
--- a/knack/arguments.py
+++ b/knack/arguments.py
@@ -10,7 +10,7 @@ from .deprecation import Deprecated
 from .preview import PreviewItem
 from .experimental import ExperimentalItem
 from .log import get_logger
-from .util import CLIError
+from .util import CLIError, status_tag_messages
 
 
 logger = get_logger(__name__)
@@ -250,8 +250,9 @@ class ArgumentsContext(object):
             return PreviewArgumentAction
 
         def _get_preview_arg_message(self):
-            return "{} '{}' is in preview. It may be changed/removed in a future release.".format(
-                self.object_type.capitalize(), self.target)
+            # "Argument xxx"
+            subject = "{} '{}'".format(self.object_type.capitalize(), self.target)
+            return status_tag_messages['preview'].format(subject)
 
         options_list = kwargs.get('options_list', None)
         object_type = 'argument'
@@ -300,8 +301,9 @@ class ArgumentsContext(object):
             return ExperimentalArgumentAction
 
         def _get_experimental_arg_message(self):
-            return "{} '{}' is experimental and not covered by customer support. " \
-                   "Please use with discretion.".format(self.object_type.capitalize(), self.target)
+            # "Argument xxx"
+            subject = "{} '{}'".format(self.object_type.capitalize(), self.target)
+            return status_tag_messages['experimental'].format(subject)
 
         options_list = kwargs.get('options_list', None)
         object_type = 'argument'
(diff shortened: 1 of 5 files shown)


knack/experimental.py
```diff
@@ -50,8 +51,7 @@ def __init__(self, cli_ctx, object_type='', target=None, tag_func=None, message_
         """
 
         def _default_get_message(self):
-            return "This {} is experimental and not covered by customer support. " \
```
jiasli, line 53: I removed "not covered by customer support", as for Knack there is no customer support at all!

knack/experimental.py
```diff
@@ -7,6 +7,7 @@
 
 _EXPERIMENTAL_TAG = '[Experimental]'
 _experimental_kwarg = 'experimental_info'
+EXPERIMENTAL_MESSAGE = "{} is experimental and under development."
```
Juliehzl, line 10: should we let customer know that experimental commands may be changed/removed in a future release?

knack/arguments.py
```diff
@@ -300,8 +302,10 @@ def __call__(self, parser, namespace, values, option_string=None):
             return ExperimentalArgumentAction
 
         def _get_experimental_arg_message(self):
-            return "{} '{}' is experimental and not covered by customer support. " \
-                   "Please use with discretion.".format(self.object_type.capitalize(), self.target)
+            from .experimental import EXPERIMENTAL_MESSAGE
```
Juliehzl, line 305: How could developers customize the message in your current design? It seems that you define the message variable with hard-coded message. If I am a developer who want to change the message, I have to re-write the message in knack. Is it by design?

knack/experimental.py
chenlomis: Hi Zunli!

Let me chime in a bit...

We actually looped in our researcher and UX writers to rephrase the messaging. Their suggestion is that _"under development_" is a more succinct expression

Also @jiasli  -- a super minor grammar call out (which I just realized)

Can we add an "is" to the sentence? --->

"{} is experimental and is under development." 

Seems to be more grammatically accurate

I hope I'm not too late with the comment...

knack/arguments.py
jiasli: A developer can customize the message like https://github.com/Azure/azure-cli/pull/15730. I will add a comment to the variable as well.

knack/experimental.py
jiasli: This message won't be seen in Azure CLI in any way. It will be overridden, like in Azure/azure-cli#15730.

> I hope I'm not too late with the comment...

Not too late at all, given @evelyn-ys is concerned about using a module variable. 

> "{} is experimental and is under development."

To me, the sentence structure is `{} is [(experimental) and (under development)]`, but I am open to this. + @dbradish-microsoft to comment.

chenlomis: Thanks Delora! @dbradish-microsoft

evelyn-ys (approved):

chenlomis (commented):
LGTM!
Zunli and Yishi may elaborate/comment better on the actual code changes

knack/util.py
```diff
@@ -11,6 +11,14 @@
 
 NO_COLOR_VARIABLE_NAME = 'KNACK_NO_COLOR'
 
+# Override these values to customize the status message.
+# The message should contain a placeholder indicating the subject (like 'This command group', 'Commend group xxx').
+# (A dict is used to avoid the "from A import B" pitfall that creates a copy of the imported B.)
```
jiasli, line 16: More info, see https://stackoverflow.com/a/3536638/2199657

PR #224, opened 2020-12-07
[Event] Add EVENT_CLI_SUCCESSFUL_EXECUTE

jiasli:
## Context

In the current implementation, the command handler is invoked at `L215`:

https://github.com/microsoft/knack/blob/0d0308bc567f2d25654bb39201c02ab262d7ce0a/knack/cli.py#L215

The JSON is printed at `L221`:

https://github.com/microsoft/knack/blob/8b3757ddff6b63bf4789c8736a506e1754f1f92b/knack/cli.py#L221

Therefore, anything printed by the command handler will be shown **before** the JSON output.

## Change

Add a new event `EVENT_CLI_SUCCESSFUL_EXECUTE`. It is raised after a command is successfully executed, allowing a callback `func` to be registered and called after JSON is printed. 

This enables post-output hint to be displayed **after** the JSON output.

### DIFF
--- a/knack/cli.py
+++ b/knack/cli.py
@@ -15,7 +15,7 @@ from .log import CLILogging, get_logger
 from .util import CLIError
 from .config import CLIConfig
 from .query import CLIQuery
-from .events import EVENT_CLI_PRE_EXECUTE, EVENT_CLI_POST_EXECUTE
+from .events import EVENT_CLI_PRE_EXECUTE, EVENT_CLI_SUCCESSFUL_EXECUTE, EVENT_CLI_POST_EXECUTE
 from .parser import CLICommandParser
 from .commands import CLICommandsLoader
 from .help import CLIHelp
@@ -219,6 +219,7 @@ class CLI(object):  # pylint: disable=too-many-instance-attributes
                 if cmd_result and cmd_result.result is not None:
                     formatter = self.output.get_formatter(output_type)
                     self.output.out(cmd_result, formatter=formatter, out_file=out_file)
+            self.raise_event(EVENT_CLI_SUCCESSFUL_EXECUTE)
         except KeyboardInterrupt as ex:
             exit_code = 1
             self.result = CommandResultItem(None, error=ex, exit_code=exit_code)
--- a/knack/cli.py
+++ b/knack/cli.py
@@ -219,7 +219,7 @@ class CLI(object):  # pylint: disable=too-many-instance-attributes
                 if cmd_result and cmd_result.result is not None:
                     formatter = self.output.get_formatter(output_type)
                     self.output.out(cmd_result, formatter=formatter, out_file=out_file)
-            self.raise_event(EVENT_CLI_SUCCESSFUL_EXECUTE)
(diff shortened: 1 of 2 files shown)


evelyn-ys (approved):

knack/cli.py
```diff
@@ -219,6 +219,7 @@ def invoke(self, args, initial_invocation_data=None, out_file=None):
                 if cmd_result and cmd_result.result is not None:
                     formatter = self.output.get_formatter(output_type)
                     self.output.out(cmd_result, formatter=formatter, out_file=out_file)
+            self.raise_event(EVENT_CLI_SUCCESSFUL_EXECUTE)
```
Juliehzl, line 222: two concerns:
1. could the new event bring benefit to CLI if exit code = 0 could represent successful execution?
2.  should SUCCESSFUL_EXECUTE be raised after cmd_result is returned? It seems that in current code output format failure would block the event.

jiasli: > could the new event bring benefit to CLI if exit code = 0 could represent successful execution?

I couldn't really get your question. Could you elaborate?

> should SUCCESSFUL_EXECUTE be raised after cmd_result is returned?

No. If `self.output.out` failed, this is considered a failure and `SUCCESSFUL_EXECUTE` shouldn't be raised by design.

knack/cli.py
```diff
@@ -219,6 +219,7 @@ def invoke(self, args, initial_invocation_data=None, out_file=None):
                 if cmd_result and cmd_result.result is not None:
                     formatter = self.output.get_formatter(output_type)
                     self.output.out(cmd_result, formatter=formatter, out_file=out_file)
+                self.raise_event(EVENT_CLI_SUCCESSFUL_EXECUTE, result=cmd_result.result)
```
fengzhou-msft, line 222: Should we raise the event only when exit_code is 0?

jiasli: Yes. On the other hand, `EVENT_CLI_POST_EXECUTE` is raised unconditionally. 

https://github.com/microsoft/knack/blob/61057a3a92756f01701c9428959a2bc1185559bf/knack/cli.py#L233-L234

fengzhou-msft: Then should you check the `exit_code` from https://github.com/microsoft/knack/blob/61057a3a92756f01701c9428959a2bc1185559bf/knack/cli.py#L217

jiasli: Not really necessary. 

https://github.com/microsoft/knack/blob/61057a3a92756f01701c9428959a2bc1185559bf/knack/cli.py#L215

will raise an exception first if `exit_code` is not `0`. 

Also,

https://github.com/microsoft/knack/blob/61057a3a92756f01701c9428959a2bc1185559bf/knack/cli.py#L221

already assumes the command is successfully executed and `exit_code` is `0`.

fengzhou-msft: got it.

fengzhou-msft (approved):

Juliehzl (approved):

PR #221, opened 2020-10-27
[Log] Print INFO log from SDK

jiasli:
1. Previously in `get_logger`, logger name is the module name prefixed with `CLI_LOGGER_NAME`, like `cli.azure.cli.core`. Now it only takes the module name, like `azure.cli.core`, to be in sync with the Python SDK's logger name `azure.*`.

2. Make `cli_logger_name` customizable, so that all `INFO` logs under `azure.*` loggers can be printed, like

   - `azure.core.pipeline.policies.http_logging_policy`
   - `azure.cli.core`

3. Previously, logger `cli.azure.cli.core`'s logger name is not printed:
   ```
   Modules found from index for 'group': ['azure.cli.command_modules.resource']
   Loading command modules:
   Name                  Load Time    Groups  Commands
   resource                  0.009        35       164
   Total (1)                 0.009        35       164
   ```
   
   Now the logger name is printed even for CLI logger when `--debug` is specified, so that the log source can be better distinguished.

ℹ As before, `DEBUG` log from both CLI and underlying libs will be printed, so this won't affect `DEBUG` log.

Example:

```log
> az group show -n fy --debug
DEBUG: knack.log: File logging enabled - writing logs to 'C:\Users\jiasli\.azure\logs'.
DEBUG: knack.cli: Command arguments: ['group', 'show', '-n', 'fy', '--debug']
DEBUG: knack.cli: Event: Cli.PreExecute []
...
DEBUG: azure.cli.core: Modules found from index for 'group': ['azure.cli.command_modules.resource']
DEBUG: azure.cli.core: Loading command modules:
DEBUG: azure.cli.core: Name                  Load Time    Groups  Commands
DEBUG: azure.cli.core: resource                  0.009        35       164
DEBUG: azure.cli.core: Total (1)                 0.009        35       164
...
DEBUG: azure.cli.core._profile: Retrieving token from ADAL for resource 'https://management.core.windows.net/'
...
DEBUG: azure.core.pipeline.policies._universal: Request URL: 'https://management.azure.com/subscriptions/0b1f6471-1bf0-4dda-aec3-cb9272f09590/resourcegroups/fy?api-version=2020-06-01'
DEBUG: azure.core.pipeline.policies._universal: Request method: 'GET'
DEBUG: azure.core.pipeline.policies._universal: Request headers:
DEBUG: azure.core.pipeline.policies._universal:     'Accept': 'application/json'
...
[...]

PR #227, opened 2020-12-28
[Log] Support multiple CLI loggers

jiasli:
Redo https://github.com/microsoft/knack/pull/221 
1. Revert `cli_logger_name` of `knack.log.CLILogging.__init__` so that no breaking change is made.
2. Support multiple cli loggers by adding more logger names to `knack.log.cli_logger_names` list.

### DIFF
--- a/knack/log.py
+++ b/knack/log.py
@@ -11,7 +11,10 @@ from .util import CtxTypeError, ensure_dir, CLIError
 from .events import EVENT_PARSER_GLOBAL_CREATE
 
 
-CLI_LOGGER_NAME = 'knack'
+CLI_LOGGER_NAME = 'cli'
+# Add more logger names to this list so that ERROR, WARNING, INFO logs from these loggers can also be displayed
+# without --debug flag.
+cli_logger_names = [CLI_LOGGER_NAME]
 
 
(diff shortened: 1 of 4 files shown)


knack/log.py
```diff
@@ -129,19 +129,23 @@ def configure(self, args):
...
-            # loggers already configured
+
+        cli_loggers = [logging.getLogger(logger_name) for logger_name in cli_logger_names]
+        for cli_logger in cli_loggers:
+            cli_logger.setLevel(logging.DEBUG)
+            cli_logger.propagate = False
+
+        if root_logger.handlers:
+            # handlers already configured
             return
```
evelyn-ys, line 145: Is it enough when `root_logger.handlers` is set? 
In previous code, both `root_logger.handlers` and `cli_logger.handlers` are checked. 
Should we check all `cli_logger.handlers for cli_logger in cli_loggers`?

jiasli: Note really necessary, as root logger and cli logger are always set at the same time. https://github.com/microsoft/knack/blob/30592e0b55e74dd1a57a73062155e44a744093c0/knack/log.py#L171-L177

evelyn-ys (approved):

PR #228, opened 2021-01-22
[Log] CLILogging.configure returns as early as possible

jiasli:
Check whether `CLILogging` already has been configured and return as early as possible, otherwise, calls like 

```py
self.log_level = self._determine_log_level(args)
```

is totally unnecessary and is merely a waste of time.

### DIFF
--- a/knack/log.py
+++ b/knack/log.py
@@ -126,11 +126,16 @@ class CLILogging:  # pylint: disable=too-many-instance-attributes
         :param args: The arguments from the command line
         :type args: list
         """
+        root_logger = logging.getLogger()
+
+        if root_logger.handlers:
+            # handlers already configured
+            return
+
         self.log_level = self._determine_log_level(args)
         console_log_levels = self._get_console_log_levels()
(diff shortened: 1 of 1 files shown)


yonzhan:
Log

evelyn-ys (approved):

fengzhou-msft (approved):

kairu-ms (approved):

knack/log.py
```diff
@@ -126,11 +126,16 @@ def configure(self, args):
         :param args: The arguments from the command line
         :type args: list
         """
+        root_logger = logging.getLogger()
+
+        if root_logger.handlers:
+            # handlers already configured
+            return
```
jiasli, line 133: When run in `pytest` environment, `root_logger` is already set by `pytest`. If we don't return here before `cli_logger.propagate = False`, `cli_logger.propagate = False` will take effect and intercepts all logs sent to `cli_loggers`.

With this change, all logs will be handled by `pytest` directly.

PR #229, opened 2021-01-25
[CI] Support Python 3.9

jiasli:
Update the pipeline to support Python 3.9.

PR #231, opened 2021-01-26
[Release] Release 0.8.0

jiasli:
Release 0.8.0

* Refine code style to comply with Python 3 (#232, #233)
* CI: Support Python 3.9 (#229)
* Logging: `CLILogging.configure` returns as early as possible (#228)


yonzhan (approved):
LGTM

fengzhou-msft (approved):

requirements.txt
```diff
@@ -1,12 +1,12 @@
...
-six==1.14.0
-tabulate==0.8.6
+argcomplete==1.12.2
+colorama==0.4.4
+flake8==3.8.3
+jmespath==0.10.0
+mock==4.0.2
+pylint==2.3.0
+Pygments==2.7.4
+PyYAML==5.4.1
```
jiasli, line 8: This `requirements.txt` is actually never used. Instead, we always honor `setup.py`.

Now sure if this will trigger issues like https://github.com/Azure/azure-cli/issues/16650, so bump them anyway.

requirements.txt
```diff
@@ -1,12 +1,12 @@
...
-tabulate==0.8.6
+argcomplete==1.12.2
+colorama==0.4.4
+flake8==3.8.4
+jmespath==0.10.0
+mock==4.0.3
+Pygments==2.7.4
+pylint==2.6.0
+pytest==6.2.2
+PyYAML==5.4.1
```
jiasli: This `requirements.txt` is actually only used by `tox`. When knack is used by Azure CLI, only `setup.py` is honored.

Now sure if this will trigger issues like https://github.com/Azure/azure-cli/issues/16650, so bump them anyway.

kairu-ms (approved):

requirements.txt
jiasli: PyYAML is unpinned in https://github.com/microsoft/knack/pull/241.

evelyn-ys (approved):

PR #235, opened 2021-03-02
Add raw_result to CommandResultItem

jiasli:
Add `raw_result` to `CommandResultItem`, so that the raw result (before query) can be used by post-output hint (https://github.com/Azure/azure-cli/pull/16242).

### DIFF
--- a/knack/cli.py
+++ b/knack/cli.py
@@ -236,7 +236,7 @@ class CLI(object):  # pylint: disable=too-many-instance-attributes
                 if cmd_result and cmd_result.result is not None:
                     formatter = self.output.get_formatter(output_type)
                     self.output.out(cmd_result, formatter=formatter, out_file=out_file)
-                self.raise_event(EVENT_CLI_SUCCESSFUL_EXECUTE, result=cmd_result.result)
+                self.raise_event(EVENT_CLI_SUCCESSFUL_EXECUTE, result=cmd_result)
         except KeyboardInterrupt as ex:
             exit_code = 1
             self.result = CommandResultItem(None, error=ex, exit_code=exit_code)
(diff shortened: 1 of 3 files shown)


knack/cli.py
```diff
@@ -236,7 +236,7 @@ def invoke(self, args, initial_invocation_data=None, out_file=None):
                 if cmd_result and cmd_result.result is not None:
                     formatter = self.output.get_formatter(output_type)
                     self.output.out(cmd_result, formatter=formatter, out_file=out_file)
-                self.raise_event(EVENT_CLI_SUCCESSFUL_EXECUTE, result=cmd_result.result)
+                self.raise_event(EVENT_CLI_SUCCESSFUL_EXECUTE, result=cmd_result)
```
jiasli, line 239: Send the full `CommandResultItem` in case post-output hint relies on other attributes:

https://github.com/microsoft/knack/blob/587abe594aa5bf1fcebd3e373f5304f285f0eb0f/knack/util.py#L23-L32

evelyn-ys: Can we add another parameter into `EVENT_CLI_SUCCESSFUL_EXECUTE` to keep `cmd_result` instead of overwrite the value of `result`? 

Although `azure-cli` didn't use `EVENT_CLI_SUCCESSFUL_EXECUTE` event now, but I'm afraid some other cli depend on knack might face breaking change in `result` type. Since event handler func usually use `func(*arg, **kwargs)` so adding new paremeter won't cause breaking change.

jiasli: Adding arg in function invocation may also trigger breaking change if the callee doesn't accept the new arg.

We never guaranteed there will be no breaking changes:

https://github.com/microsoft/knack

> The project is in initial development phase. We recommend pinning to at least a specific minor version when marking knack as a dependency in your project.

evelyn-ys (approved):
Ok with me if breaking change is acceptable😆

PR #238, opened 2021-03-10
Init colorama only in Windows legacy terminal

jiasli:
Init colomara only on Windows, because

- Bash of Linux and MacOS already supports control sequences
- Removing the colomara wrapper increases the output performance
- `_should_init_colorama` can be furtherly disabled by Azure CLI

### DIFF
--- a/knack/cli.py
+++ b/knack/cli.py
@@ -21,10 +21,6 @@ from .help import CLIHelp
 
 logger = get_logger(__name__)
 
-# Temporarily force color to be enabled even when out_file is not stdout.
-# This is only intended for testing purpose.
-_KNACK_TEST_FORCE_ENABLE_COLOR = False
-
 
 class CLI(object):  # pylint: disable=too-many-instance-attributes
     """ The main driver for the CLI """
@@ -209,7 +205,7 @@ class CLI(object):  # pylint: disable=too-many-instance-attributes
         exit_code = 0
         try:
             out_file = out_file or self.out_file
-            if out_file is sys.stdout and self._should_init_colorama or _KNACK_TEST_FORCE_ENABLE_COLOR:
+            if out_file is sys.stdout and self._should_init_colorama:
                 self.init_debug_log.append("Init colorama.")
                 import colorama
                 colorama.init()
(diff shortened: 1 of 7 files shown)


knack/cli.py
```diff
@@ -101,6 +101,8 @@ def __init__(self,
 
         self.only_show_errors = self.config.getboolean('core', 'only_show_errors', fallback=False)
         self.enable_color = self._should_enable_color()
+        # Init colorama only in Windows legacy terminal
+        self._should_init_colorama = self.enable_color and os.name == 'nt' and not is_modern_terminal()
```
evelyn-ys, line 101: Why not `(self.enable_color or _KNACK_TEST_FORCE_ENABLE_COLOR) and ...`? Does that mean knack test always use colorama no matter it's modern terminal or legacy terminal?

jiasli: > Why not `(self.enable_color or _KNACK_TEST_FORCE_ENABLE_COLOR) and ...`? 

The action scope of mocked `_KNACK_TEST_FORCE_ENABLE_COLOR` is `CLI.invoke`. If `_KNACK_TEST_FORCE_ENABLE_COLOR` is used in `CLI.__invoke__`, `CLI.invoke` won't be affected and won't have colorama enabled.

https://github.com/microsoft/knack/blob/cd590312b9b4ea1402d7a0a76814f6ca4da7fab8/tests/util.py#L36-L37

> Does that mean knack test always use colorama no matter it's modern terminal or legacy terminal?

Yes.

jiasli: Dropped `_KNACK_TEST_FORCE_ENABLE_COLOR`. Now strip control sequences by knack itself:

https://github.com/microsoft/knack/blob/e5e0529db43a2dbef50096237d82d3003506a34f/tests/util.py#L57-L58

tests/util.py
```diff
@@ -54,8 +54,17 @@ def wrapper(self):
...
+    return re.sub(r'\x1b[^m]+m', '', string)
+
+
+def _remove_whitespace(string):
+    return re.sub(r'\s', '', string)
+
+
+def assert_in_multi_line(sub_string, string):
+    # assert sub_string is in string, with all whitespaces, line breaks and control sequences ignored
+    assert _remove_whitespace(sub_string) in _remove_control_sequence(_remove_whitespace(string))
```
jiasli, line 67: This function does the job of removing control sequences which are previously removed by colorama.

evelyn-ys (approved):

knack/cli.py
evelyn-ys: huge work😂

fengzhou-msft (approved):

PR #178, opened 2020-03-03
Show warning when UnicodeEncodeError is raised

jiasli:
Show warning when `UnicodeEncodeError` is raised.

To test, on Windows go to **Control Panel > Region > Administrative > Change system local** and use English. Do not check **Beta: Use Unicode UTF-8 for worldwide language support**.

Open PowerShell terminal and run

```jsonc
az group show -n testrg
{
  "id": "/subscriptions/0b1f6471-1bf0-4dda-aec3-cb9272f09590/resourceGroups/testrg",
  "location": "centralus",
  "managedBy": null,
  "name": "testrg",
  "properties": {
    "provisioningState": "Succeeded"
  },
  "tags": {
    "key0303": "测试€£￥テスト안녕𓀀𓀁𒀉𒅤"
  },
  "type": "Microsoft.Resources/resourceGroups"
}
```

The output is correct, but if the output is redirected, PowerShell uses the system encoding. Thus `az` will silently remove the value of `key0303`:

```jsonc
> $t = az group show -n testrg
> $t
{
  "id": "/subscriptions/0b1f6471-1bf0-4dda-aec3-cb9272f09590/resourceGroups/testrg",
  "location": "centralus",
  "managedBy": null,
  "name": "testrg",
  "properties": {
    "provisioningState": "Succeeded"
  },
  "tags": {
    "key0303": ""   // value is removed
  },
  "type": "Microsoft.Resources/resourceGroups"
}
```

The solution is any of 

1. Remove invalid characters under the current locale
2. Use a locale that supports those special characters
3. Check **Beta: Use Unicode UTF-8 for worldwide language support**

utf8

If none of the above solution is adopted, Knack will show a warning:

```
> $t = az group show -n testrg
Unable to encode the output with cp1252 encoding. Unsupported characters are discarded.**
```

PR #247, opened 2021-05-07
Always use UTF-8 for log file encoding

jiasli:
## Context

Reported by https://github.com/Azure/azure-cli/issues/17994

On Windows with English as the system language (without UTC-8 enabled), the system encoding by default is `cp1252` (Western Europe), and Python will use `cp1252` as the file encoding by default. 

Writing Unicode characters like `汉字` to log file results in error:

```
Traceback (most recent call last):
  File "C:\Users\xxx\AppData\Local\Programs\Python\Python38\lib\logging\__init__.py", line 1084, in emit
    stream.write(msg + self.terminator)
  File "C:\Users\xxx\AppData\Local\Programs\Python\Python38\lib\encodings\cp1252.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
UnicodeEncodeError: 'charmap' codec can't encode characters in position 87-88: character maps to <undefined>
```

The error can also be easily reproduced with 

```py
with open("test.txt", "w") as f:
    print(f.encoding)
    f.write("汉字")
```

```
cp1252
Traceback (most recent call last):
  File "D:/cli/testproj/test1.py", line 2, in <module>
    f.write("汉字")
  File "C:\Users\jiasli\AppData\Local\Programs\Python\Python38\lib\encodings\cp1252.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
UnicodeEncodeError: 'charmap' codec can't encode characters in position 0-1: character maps to <undefined>
```

## Change

This PR forces log files to use UTF-8 so that file logging works on non-UTF-8 systems as well.

## Alternative solution

One may also follow https://github.com/microsoft/knack/pull/178 to change the default encoding of the system to UTF-8.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,6 +3,11 @@
 Release History
 ===============
 
+0.8.2
++++++
+
+* Always use UTF-8 for log file encoding (#247)
+
 0.8.1
 +++++
 
--- a/knack/log.py
+++ b/knack/log.py
@@ -16,6 +16,8 @@ CLI_LOGGER_NAME = 'cli'
 # without --debug flag.
 cli_logger_names = [CLI_LOGGER_NAME]
 
+LOG_FILE_ENCODING = 'utf-8'
+
 
 class CliLogLevel(IntEnum):
     CRITICAL = 0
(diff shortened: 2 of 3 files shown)


jiasli:
If you want to test, simply enable file logging with `az config set logging.enable_log_file=true` and write some Chinese charaters to the log:

```py
logger.warning("汉字")
```

yonzhan (approved):
LGTM

evelyn-ys (approved):

avanigupta (approved):

PR #248, opened 2021-06-02
Always use unittest.mock rather than PyPI mock.

musicinmybrain:
Since only Python 3.6+ is supported, there is no need to use the `mock`
backport package. This commit ensures that only the standard library’s
`unittest.mock` is needed.

Replaces a mixture of conditional and unconditional imports of `mock`
and/or `unittest.mock` with unconditional imports of `unittest.mock`,
and drops `mock` from `requirements.txt`.

### DIFF
--- a/knack/testsdk/patches.py
+++ b/knack/testsdk/patches.py
@@ -4,10 +4,7 @@
 # --------------------------------------------------------------------------------------------
 
 import unittest
-try:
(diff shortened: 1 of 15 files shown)


jiasli:
Thanks for the contribution! Nice change!

jiasli (approved):

PR #233, opened 2021-01-27
Drop Python 2 compatible code

jiasli:
Drop Python 2 compatible code by:

- removing `six` (same as https://github.com/Azure/azure-cli/pull/16082)
- removing `from __future__ import print_function`

PR #249, opened 2021-09-17
Only install colorama on Windows

jiasli:
As colorama is only initialized on Windows (#238), it should not be installed on Linux or MacOS.

### DIFF
--- a/knack/cli.py
+++ b/knack/cli.py
@@ -98,7 +98,7 @@ class CLI(object):  # pylint: disable=too-many-instance-attributes
         self.only_show_errors = self.config.getboolean('core', 'only_show_errors', fallback=False)
         self.enable_color = self._should_enable_color()
         # Init colorama only in Windows legacy terminal
-        self._should_init_colorama = self.enable_color and os.name == 'nt' and not is_modern_terminal()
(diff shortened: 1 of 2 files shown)


knack/cli.py
```diff
@@ -98,7 +98,7 @@ def __init__(self,
         self.only_show_errors = self.config.getboolean('core', 'only_show_errors', fallback=False)
         self.enable_color = self._should_enable_color()
         # Init colorama only in Windows legacy terminal
-        self._should_init_colorama = self.enable_color and os.name == 'nt' and not is_modern_terminal()
+        self._should_init_colorama = self.enable_color and sys.platform == 'win32' and not is_modern_terminal()
```
jiasli, line 101: Use the same method as `setup.py` to detect system.

setup.py
```diff
@@ -5,21 +5,24 @@
 # Licensed under the MIT License. See License.txt in the project root for license information.
 # --------------------------------------------------------------------------------------------
 
-from codecs import open
```
jiasli, line 8: This is for Python 2 which has been dropped long time ago (https://github.com/microsoft/knack/pull/233).

evelyn-ys (approved):

PR #253, opened 2021-11-12
Fix `--only-show-errors` on root parser

jiasli:
`--only-show-errors` doesn't work on root parser:

```
> python .\examples\exapp2 --only-show-errors
usage: exapp2 [-h] [--verbose] [--debug] [--only-show-errors] [--output {json,jsonc,yaml,yamlc,table,tsv,none}]
              [--query JMESPATH]
              {hello,sample-json,sample-logger,abc,ga,pre,exp,dep,demo} ...
exapp2: error: the following arguments are required: _command
```

This is due to a typo made by #179. Now fix it.

### DIFF
--- a/knack/invocation.py
+++ b/knack/invocation.py
@@ -137,7 +137,7 @@ class CommandInvoker(object):
         self.parser.load_command_table(self.commands_loader)
         self.cli_ctx.raise_event(EVENT_INVOKER_CMD_TBL_LOADED, parser=self.parser)
 
-        arg_check = [a for a in args if a not in ['--verbose', '--debug', '--only-show-warnings']]
+        arg_check = [a for a in args if a not in ['--verbose', '--debug', '--only-show-errors']]
         if not arg_check:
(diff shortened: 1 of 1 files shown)


kairu-ms (approved):

knack/invocation.py
```diff
@@ -137,7 +137,7 @@ def execute(self, args):
         self.parser.load_command_table(self.commands_loader)
         self.cli_ctx.raise_event(EVENT_INVOKER_CMD_TBL_LOADED, parser=self.parser)
 
-        arg_check = [a for a in args if a not in ['--verbose', '--debug', '--only-show-warnings']]
```
kairu-ms, line 140: It's overwritten in azure-cli-core

https://github.com/Azure/azure-cli/blob/dce83e308546a3fda4085055b35d7b69fff43b33/src/azure-cli-core/azure/cli/core/commands/__init__.py#L565

Issue #256, opened 2022-03-01
Colorama is not being installed in Windows even though it is required


niander:
Since version 0.8.3 (which has been removed from public feeds) `colorama` is only added to the list of requirements if the setup.py is executed in a Windows machine. However, knack is being distributed as a wheel package and setup.py is not executed when installing a wheel package. Therefore, knack is broken when executed from a terminal.

conditional dependency needs to follow [PEP 508](https://www.python.org/dev/peps/pep-0508/).

PR #257, opened 2022-03-01
Use proper PEP 508 environment marker to install colorama on Windows only

niander:
Issue #256

Using proper PEP 508 environment marker to install colorama on Windows only.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,7 +8,7 @@
 import sys
 from setuptools import setup
 
-VERSION = '0.9.0'
+VERSION = '0.9.1'
 
 DEPENDENCIES = [
     'argcomplete',
--- a/setup.py
+++ b/setup.py
@@ -15,12 +15,10 @@ DEPENDENCIES = [
     'jmespath',
     'pygments',
     'pyyaml',
-    'tabulate'
+    'tabulate',
+    # On Windows, colorama is required for legacy terminals.
+    'colorama; sys_platform == "win32"'
 ]
-
-# On Windows, colorama is required for legacy terminals.
-if sys.platform == 'win32':
-    DEPENDENCIES.append('colorama')
 
 with open('README.rst', 'r') as f:
     README = f.read()
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,6 +3,11 @@
 Release History
 ===============
 
+0.9.1
++++++
+
+* Use proper PEP 508 environment marker to install colorama on Windows only (#257)
+
 0.9.0
 +++++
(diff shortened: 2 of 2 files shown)


jiasli:
Indeed, after decompressing https://files.pythonhosted.org/packages/bf/9e/6378bc3d04592cd3692dd50f829c6678fae69052fda048a7981657c17489/knack-0.9.0-py3-none-any.whl, `knack-0.9.0.dist-info/METADATA` contains:

```
Requires-Dist: argcomplete
Requires-Dist: jmespath
Requires-Dist: pygments
Requires-Dist: pyyaml
Requires-Dist: tabulate
```

setup.py
```diff
@@ -8,20 +8,18 @@
...
 
 DEPENDENCIES = [
     'argcomplete',
     'jmespath',
     'pygments',
     'pyyaml',
-    'tabulate'
+    'tabulate',
+    # On Windows, colorama is required for legacy terminals.
+    'colorama; sys_platform == "win32"'
```
jiasli, line 20: I am actually wondering which one is the best:

- [`os_name`](https://docs.python.org/3/library/os.html#os.name)
- [`sys_platform`](https://docs.python.org/3/library/sys.html#sys.platform)
- [`platform_system`](https://docs.python.org/3/library/platform.html#platform.system)

`portalocker` is one dependency of Azure CLI and it uses:

https://github.com/WoLpH/portalocker/blob/00cffcb1831208691d3bf6850df4768989a0bd4c/setup.py#L133

```py
            'pywin32>=226; platform_system == "Windows"',
```

jiasli (approved):

calvinhzy (approved):

setup.py
niander: I have seen several projects to use `platform_system == "Windows"` before, but none to use `sys_platform`. I don't think it matters but maybe we should use `platform_system` because it is more commonly used?

niander: Oh, I see this has already been merged so I guess it doesn't matter. But.. Would ever be the case `sys_platform` would be something else than `win32` in a Windows? Like `win64`?

jiasli: > Would ever be the case `sys_platform` would be something else than `win32` in a Windows? Like `win64`?

I don't think so. I am running 64-bit Python on 64-bit Windows:

```
> python -c "import sys; print(sys.platform)"
win32
```

https://docs.python.org/3/library/sys.html#sys.platform only mentions `win32` too.

niander:
When will this be released?

PR #265, opened 2022-08-15
Enable Virtual Terminal mode on legacy Windows terminal to support ANSI escape sequences

jiasli:
Migrated from https://github.com/Azure/azure-cli/pull/12942

This PRs totally removes colorama, getting us rid of colorama bugs.

Windows versions before Windows 10 1607 are not supported anymore. Colors are automatically disabled on those jurassic period terminals 🦖.

When [Windows Terminal](https://github.com/microsoft/terminal) becomes universal, even `_win_vt.py` should be dropped.

### DIFF
--- a/knack/cli.py
+++ b/knack/cli.py
@@ -97,8 +97,8 @@ class CLI(object):  # pylint: disable=too-many-instance-attributes
 
         self.only_show_errors = self.config.getboolean('core', 'only_show_errors', fallback=False)
         self.enable_color = self._should_enable_color()
-        # Init colorama only in Windows legacy terminal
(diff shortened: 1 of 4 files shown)


knack/cli.py
```diff
@@ -249,10 +251,6 @@ def invoke(self, args, initial_invocation_data=None, out_file=None):
         finally:
             self.raise_event(EVENT_CLI_POST_EXECUTE)
 
-            if self._should_init_colorama:
-                import colorama
-                colorama.deinit()
-
```
jiasli, line 255: I don't see any side effect if we don't disable it.

evelyn-ys (approved):

zhoxing-ms (approved):

PR #273, opened 2023-07-19
fix default value set for bool type

AllyW:
`bool` is subclass of int, while default `True` or `False` should not be converted into int `1` or `0`.
Otherwise, type check would fail for `aaz` cmds (https://github.com/Azure/azure-cli/blob/58ac6ab07fa850e1a6458dec87af91cd6725069e/src/azure-cli-core/azure/cli/core/aaz/_field_type.py#L47)

### DIFF
--- a/knack/commands.py
+++ b/knack/commands.py
@@ -124,8 +124,8 @@ class CLICommand(object):  # pylint:disable=too-many-instance-attributes
         # that coincides with the default
         if isinstance(arg_default, str):
             arg_default = DefaultStr(arg_default)
-        elif type(arg_default) is int:  # pylint: disable=unidiomatic-typecheck
-            # use type here is because:
+        elif isinstance(arg_default, int) and not isinstance(arg_default, bool):
+            # adjust here is because:
             # 1) bool is subclass of int so isinstance(True, int) cannot distinguish between int and bool
             # 2) bool is not extendable according to
             #   https://stackoverflow.com/questions/2172189/why-i-cant-extend-bool-in-python,
--- a/knack/commands.py
+++ b/knack/commands.py
@@ -17,7 +17,7 @@ from .introspection import extract_args_from_signature, extract_full_summary_fro
 from .events import (EVENT_CMDLOADER_LOAD_COMMAND_TABLE, EVENT_CMDLOADER_LOAD_ARGUMENTS,
                      EVENT_COMMAND_CANCELLED)
 from .log import get_logger
-from .validators import DefaultInt, DefaultStr
+from .validators import DefaultInt, DefaultStr, DefaultBool
 
 logger = get_logger(__name__)
 
@@ -125,7 +125,7 @@ class CLICommand(object):  # pylint:disable=too-many-instance-attributes
         if isinstance(arg_default, str):
             arg_default = DefaultStr(arg_default)
         elif isinstance(arg_default, bool):
-            pass
+            arg_default = DefaultBool(arg_default)
         elif isinstance(arg_default, int):
             arg_default = DefaultInt(arg_default)
         # update the default
--- a/knack/commands.py
+++ b/knack/commands.py
@@ -129,7 +129,7 @@ class CLICommand(object):  # pylint:disable=too-many-instance-attributes
         elif isinstance(arg_default, int):
             arg_default = DefaultInt(arg_default)
         # update the default
-        if arg_default:
+        if arg_default is not None:
             arg.type.settings['default'] = arg_default
 
     def execute(self, **kwargs):
--- a/knack/commands.py
+++ b/knack/commands.py
@@ -124,6 +124,8 @@ class CLICommand(object):  # pylint:disable=too-many-instance-attributes
         # that coincides with the default
         if isinstance(arg_default, str):
             arg_default = DefaultStr(arg_default)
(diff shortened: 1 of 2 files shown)


AllyW:
in this pr: https://github.com/Azure/azure-cli/pull/26819
aaz cmd:
image

before:
image
after:
image

knack/commands.py
```diff
@@ -124,6 +124,8 @@ def update_argument(self, param_name, argtype):
         # that coincides with the default
         if isinstance(arg_default, str):
             arg_default = DefaultStr(arg_default)
+        elif isinstance(arg_default, bool):
+            pass
```
jiasli: I don't think we can simply `pass` here, as this will make it impossible to distinguish whether the received `True` is user specified or the default value.

kairu-ms: Please add comments in the code here.

kairu-ms: > I don't think we can simply `pass` here, as this will make it impossible to distinguish whether the received `True` is user specified or the default value.

It only pass the converting of bool to int. Not pass the default value

kairu-ms: I think it's useless, because it only added `is_default` property for str and int, not cover the other types, such as bool, float, timestamp etc.

kairu-ms: BTW, I also found the following lines has a bug when the arg_default is `0` int value https://github.com/microsoft/knack/blob/e496c9590792572e680cb3ec959db175d9ba85dd/knack/commands.py#L129-L131

AllyW:
corresponding swagger line：
https://github.com/Azure/azure-rest-api-specs/blob/493aa7224fd65fe1e5b4cff59bcae5c6cdf4525e/specification/monitor/resource-manager/Microsoft.Insights/preview/2023-05-01-preview/tenantActionGroups_API.json#L570

knack/commands.py
jiasli: Could you please elaborate what the bug is?

jiasli:
Also see https://stackoverflow.com/questions/37888620/comparing-boolean-and-int-using-isinstance

knack/commands.py
```diff
@@ -124,7 +124,12 @@ def update_argument(self, param_name, argtype):
         # that coincides with the default
         if isinstance(arg_default, str):
             arg_default = DefaultStr(arg_default)
-        elif isinstance(arg_default, int):
+        elif type(arg_default) is int:  # pylint: disable=unidiomatic-typecheck
```
kairu-ms: Why not state it like this, it's much clear
```suggestion
        elif isinstance(arg_default, int) and not isinstance(arg_default, bool):
```

kairu-ms (approved):

knack/commands.py
```diff
@@ -124,7 +124,12 @@ def update_argument(self, param_name, argtype):
         # that coincides with the default
         if isinstance(arg_default, str):
             arg_default = DefaultStr(arg_default)
-        elif isinstance(arg_default, int):
+        elif isinstance(arg_default, int) and not isinstance(arg_default, bool):
+            # adjust here is because:
+            # 1) bool is subclass of int so isinstance(True, int) cannot distinguish between int and bool
+            # 2) bool is not extendable according to
+            #   https://stackoverflow.com/questions/2172189/why-i-cant-extend-bool-in-python,
+            #   so bool's is_default is ignored for now
```
jiasli: Code comment should usually not focus on adjustment, but the status quo.

We can get the short link for a Stackoverflow post via the "share" button:

image

```suggestion
            # bool's is_default is ignored for now, because
            # 1. bool is a subclass of int, so isinstance(True, int) cannot distinguish between int and bool
            # 2. bool is not extendable according to https://stackoverflow.com/a/2172204/2199657
```

knack/commands.py
jiasli: After reviewing the code, I feel this `if arg_default:` statement is deliberately added in order to ignore `False`/`0` as default.

For example, for arguments `--disable-integrity-monitoring` and `--disable-integrity-monitoring-autoupgrade` of `az vm create`, their default values are `False`/`0`, so they are deliberately omitted from the help message:

```
> az vm create -h
...
    --disable-integrity-monitoring                      : Disable the default behavior of installing
                                                          guest attestation extension and enabling
                                                          System Assigned Identity for Trusted
                                                          Launch enabled VMs and VMSS.
    --disable-integrity-monitoring-autoupgrade          : Disable auto upgrade of guest attestation
                                                          extension for Trusted Launch enabled VMs
                                                          and VMSS.
```

jiasli (approved):

jiasli:
As a side note, https://github.com/Azure/azure-cli/pull/16081 took a different approach. It sets the default value of `role` to `None` and defaulting it to `Contributor` manually in `custom.py`'s code.

jiasli:
The C source code of `bool` also mentions:

https://github.com/python/cpython/blob/832c37d42a395d4ea45994daffa5e41bd74ac1bb/Objects/boolobject.c#L118

> The class bool is a subclass of the class int, and cannot be subclassed.

An alternative is to consider developing our own `Bool` and implementing special methods such as `__bool__`, `__str__`, so that it acts like a real `bool`, and also bears additional attribute `is_default`.

evelyn-ys (approved):

bebound (approved):

PR #275, opened 2023-07-21
Declare support for Python 3.11 and drop support for Python 3.7

jiasli:
Declare support for Python 3.11 and drop support for Python 3.7.

Corresponding Azure CLI issues:

- https://github.com/Azure/azure-cli/issues/26537
- https://github.com/Azure/azure-cli/issues/24494


knack/testsdk/base.py
```diff
@@ -117,7 +117,7 @@ def setUp(self):
 
         # set up cassette
         cm = self.vcr.use_cassette(self.recording_file)
-        self.cassette = cm.__enter__()
+        self.cassette = cm.__enter__()  # pylint: disable=unnecessary-dunder-call
```
jiasli, line 120: `cm.__exit__` is not called immediately in this method to release `self.cassette`, but registered in `self.addCleanup`, so this dunder call is not unnecessary.

knack/introspection.py
```diff
@@ -67,45 +67,23 @@ def extract_args_from_signature(operation, excluded_params=None):
...
         action = 'store_' + str(not default).lower() if isinstance(default, bool) else None
-
-        try:
-            default = (default
-                       if default != inspect._empty  # pylint: disable=protected-access
-                       else None)
-        except AttributeError:
-            pass
-
+        command_argument_default = default if default != inspect.Parameter.empty else None
```
jiasli, line 80: The original code reuses the variable name `default` to denote different things, which is a bad practice. We rename it to `command_argument_default` to distinguish **command argument default** from **function argument default**.

knack/introspection.py
```diff
@@ -67,45 +67,23 @@ def extract_args_from_signature(operation, excluded_params=None):
...
-    args = []
-    try:
-        # only supported in python3 - falling back to argspec if not available
-        sig = inspect.signature(operation)
-        args = sig.parameters
-    except AttributeError:
-        sig = inspect.getargspec(operation)  # pylint: disable=deprecated-method, useless-suppression
-        args = sig.args
+    sig = inspect.signature(operation)
+    args = sig.parameters
```
jiasli, line 71: [`inspect.getargspec`](https://docs.python.org/3.10/library/inspect.html#inspect.getargspec) is removed in Python 3.11 and replaced by [`inspect.getfullargspec`](https://docs.python.org/3.11/library/inspect.html#inspect.getfullargspec). However, `inspect.getfullargspec` is retained only to maintain backward compatibility. The recommendation is to use [`inspect.signature`](https://docs.python.org/3.11/library/inspect.html#inspect.signature):

> Note that [signature()](https://docs.python.org/3.11/library/inspect.html#inspect.signature) and [Signature Object](https://docs.python.org/3.11/library/inspect.html#inspect-signature-object) provide the recommended API for callable introspection, and support additional behaviours (like positional-only arguments) that are sometimes encountered in extension module APIs. This function is retained primarily for use in code that needs to maintain compatibility with the Python 2 inspect module API.
> -- https://docs.python.org/3.11/library/inspect.html#inspect.getfullargspec

Given we have dropped Python 2 long ago (https://github.com/microsoft/knack/pull/233), it's time to drop the usage of `inspect.getargspec` too.

azure-pipeline.yml
```diff
@@ -53,9 +53,8 @@ jobs:
       name: 'pool-ubuntu-2004'
     steps:
       - task: UsePythonVersion@0
-        displayName: Use Python 3.10
```
jiasli, line 56: Using the default `displayName` is sufficient - it is the same as what we are specifying here.

image

tox.ini
```diff
@@ -1,10 +1,10 @@
 [tox]
-envlist = py36,py37,py38,py39,py310
+envlist = py38,py39,py310,py311
 [testenv]
 deps = -rrequirements.txt
 commands=
     python ./scripts/license_verify.py
     flake8 --statistics --append-config=.flake8 knack
-    pylint knack --rcfile=.pylintrc -r n -d I0013
+    pylint knack --rcfile=.pylintrc --reports n --disable I0013
```
jiasli, line 8: Use arguments' full name for clarity.

It's very likely we need to remove `--disable I0013` in the near future, as pylint 3.0.0b1 disables it by default:

> This message is disabled by default. To enable it, add file-ignored to the enable option.
> -- https://pylint.pycqa.org/en/latest/user_guide/messages/information/file-ignored.html

knack/deprecation.py
```diff
@@ -94,12 +94,10 @@ def _default_get_message(self):
...
         )
 
-    # pylint: disable=no-self-use
     def _version_less_than_or_equal_to(self, v1, v2):
         """ Returns true if v1 <= v2. """
-        # pylint: disable=no-name-in-module, import-error
-        from distutils.version import LooseVersion
-        return LooseVersion(v1) <= LooseVersion(v2)
+        from packaging.version import parse
+        return parse(v1) <= parse(v2)
```
jiasli, line 100: Similar changes were previously made in https://github.com/Azure/azure-cli/pull/17667.

[PEP 632](https://peps.python.org/pep-0632/#migration-advice) mentions:

> `distutils.version` — use the `packaging` package

knack/cli.py
```diff
@@ -104,15 +104,15 @@ def __init__(self,
     def _should_show_version(args):
         return args and (args[0] == '--version' or args[0] == '-v')
 
-    def get_cli_version(self):  # pylint: disable=no-self-use
+    def get_cli_version(self):
```
jiasli, line 107: The removal of `# pylint: disable=no-self-use` is explained in https://github.com/Azure/azure-cli-dev-tools/pull/359#discussion_r1259334166

evelyn-ys (approved):

bebound (approved):

wangzelin007 (approved):

PR #279, opened 2023-10-06
Mark Python 3.12 support

tonybaloney:
I've run the tests locally and all pass

### DIFF
--- a/azure-pipeline.yml
+++ b/azure-pipeline.yml
@@ -32,6 +32,9 @@ jobs:
         Python311:
           python.version: '3.11'
           tox_env: 'py311'
+        Python312:
+          python.version: '3.12'
+          tox_env: 'py312'
     steps:
       - task: UsePythonVersion@0
         displayName: 'Use Python $(python.version)'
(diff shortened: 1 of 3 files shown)


onlined:
@jiasli @yonzhan @AllyW sorry for mentioning directly, but can someone review this? It's a simple and small PR.

jiasli:
We will work on Python 3.12 support together with https://github.com/Azure/azure-cli/issues/27673.

onlined:
I think this can be merged independently, I don't understand, are these projects coupled? Or don't you trust the tests enough that you can't merge before you run Azure CLI tests with Knack on Python 3.12?

jiasli:
There is much more to do than declaring 3.12 support. See #275 as an example.

azure-pipeline.yml
```diff
@@ -32,6 +32,9 @@ jobs:
         Python311:
           python.version: '3.11'
           tox_env: 'py311'
+        Python312:
+          python.version: '3.12'
+          tox_env: 'py312'
```
jiasli, line 37: Could you also update L60 to use `3.12`?

evelyn-ys (approved):

bebound (approved):

jiasli (approved):

azure-pipeline.yml
jiasli: I will make the change in #285.

PR #284, opened 2024-07-10
Drop .travis.yml

jiasli:
We have switched to Azure Pipelines long ago. The Travis pipeline file `.travis.yml` is no longer maintained.


bebound (approved):

.travis.yml
```diff
@@ -1,12 +0,0 @@
-arch:
-- amd64
-- ppc64le
-sudo: false
-language: python
-python:
-- '3.8'
-- '3.9'
-- '3.10'
-- '3.11'
```
jiasli, line 10: If we don't drop this file, we have to maintain these lines which is unnecessary.

evelyn-ys (approved):

PR #254, opened 2021-11-12
Use Python 3.10 for BuildPythonWheel

jiasli:
Use Python 3.10 for BuildPythonWheel

PR #296, opened 2026-04-01
Mark Python 3.14 support and drop Python 3.9

jiasli:
Mark Python 3.14 support and drop Python 3.9, following the same pattern as:

- #279
- #290

---

🤖 This PR is assisted by VS Code GitHub Copilot, with Claude Opus 4.6 Medium model, and human-verified.

Prompt:

> Follow the same pattern of https://github.com/microsoft/knack/pull/279 and https://github.com/microsoft/knack/pull/290 to add Python 3.14 support and meanwhile remove Python 3.9 support.


azure-pipeline.yml
```diff
@@ -57,7 +57,7 @@ jobs:
     steps:
       - task: UsePythonVersion@0
         inputs:
-          versionSpec: 3.13
+          versionSpec: '3.14'
```
jiasli, line 60: Quote the version number according to https://github.com/microsoft/knack/pull/296#discussion_r3020458280.

Unquoted `3.10` did appear before in https://github.com/microsoft/knack/pull/254, but it didn't cause any issue. Here is the explanation from Copilot:

---

You're right — that commit does show `versionSpec: 3.10` unquoted. So the question is valid: why didn't that break?

The answer is that **Azure Pipelines doesn't use a strict YAML 1.1 parser**. Azure Pipelines uses a custom YAML parser that treats values like `3.10` as strings, not floats. So `3.10` is not silently truncated to `3.1`.

This is different from standard YAML 1.1 (used by tools like PyYAML), where `3.10` would indeed be parsed as the float `3.1`. But Azure Pipelines specifically handles version-like values as strings to avoid this exact pitfall.

So the Copilot review comment is **technically correct in general YAML terms** (quoting is best practice and safer), but **practically incorrect** — it wouldn't actually cause an issue in Azure Pipelines. The existing unquoted `3.10` worked fine for the same reason: Azure Pipelines doesn't apply float coercion to these values.

That said, quoting is still good hygiene for portability and clarity.

bebound (approved):

PR #300, opened 2026-04-29
{Misc.} Fix help `%` token handling for Python 3.14 argparse compatibility

YangAn-microsoft:
### Summary
This PR fixes Knack help rendering for `%`-based help strings to keep behavior stable with Python 3.14 argparse validation. Replaces #299 and optionally link any key discussion from #299.

### Why this is needed (Python 3.14)
Python 3.14 argparse performs stricter help-format validation.  
Help text containing unescaped literal `%` can raise formatting errors during help processing.  
Since Knack renders help through custom paths, we must handle these cases safely while preserving argparse placeholder semantics.

### What changed
1. Added safe `%` sanitization for argparse-facing help text.
2. Expanded argparse placeholders in Knack-rendered help output.
3. Added safe fallback behavior so malformed/literal `%` does not crash help rendering.
4. Added/updated tests for:
   - `%(default)s`
   - `%(prog)s`
   - `%%`
   - single `%` fallback behavior

### Resulting behavior
1. `%(default)s` expands to the argument default value.
2. `%(prog)s` expands to the parser program string.
3. `%%` is rendered as `%`.
4. single `%` does not crash and is shown as `%`.

### Validation
1. Targeted token tests passed.
2. Full Knack test suite passed on Python 3.14 (`245 passed`).
3. Azure CLI help-path checks with local Knack changes completed without parser/help crashes.

### Scope
This is a focused compatibility and robustness fix for help rendering behavior under Python 3.14.

### DIFF
--- a/knack/parser.py
+++ b/knack/parser.py
@@ -34,47 +34,18 @@ class CLICommandParser(argparse.ArgumentParser):
 
     @staticmethod
     def _sanitize_help_for_argparse(help_text):
-        """Escape literal '%' while preserving argparse mapping placeholders.
+        """Escape literal '%' for argparse unless the author already escaped.
 
-        argparse interpolates help text with ``%`` formatting against a dict.
-        Keep valid ``%(name)s``-style placeholders as-is and escape everything
-        else so help text like date formats (for example, ``%Y-%m-%d``) doesn't
-        crash during parser construction.
+        If a help string already contains ``%(...)`` placeholders or ``%%``,
+        keep it as-is and rely on argparse's native formatting behavior.
+        Otherwise, escape ``%`` so literal percent tokens don't break help
+        processing in newer Python versions.
         """
         if not isinstance(help_text, str) or '%' not in help_text:
             return help_text
-
-        result = []
-        idx = 0
-        text_len = len(help_text)
-        while idx < text_len:
-            char = help_text[idx]
-            if char != '%':
-                result.append(char)
-                idx += 1
-                continue
-
-            # Keep already-escaped percent signs as-is.
-            if idx + 1 < text_len and help_text[idx + 1] == '%':
-                result.append('%%')
-                idx += 2
-                continue
-
-            # Preserve mapping placeholders, e.g. %(default)s.
-            if idx + 1 < text_len and help_text[idx + 1] == '(':
-                closing_paren = help_text.find(')', idx + 2)
-                if closing_paren != -1 and closing_paren + 1 < text_len:
-                    conversion_char = help_text[closing_paren + 1]
-                    if conversion_char.isalpha():
-                        result.append(help_text[idx:closing_paren + 2])
-                        idx = closing_paren + 2
-                        continue
-
-            # Any other '%' is literal and must be escaped for argparse.
-            result.append('%%')
-            idx += 1
-
-        return ''.join(result)
+        if '%(' in help_text or '%%' in help_text:
+            return help_text
+        return help_text.replace('%', '%%')
 
     @staticmethod
     def create_global_parser(cli_ctx=None):
--- a/knack/parser.py
+++ b/knack/parser.py
@@ -33,6 +33,50 @@ ARGPARSE_SUPPORTED_KWARGS = [
 class CLICommandParser(argparse.ArgumentParser):
 
     @staticmethod
+    def _sanitize_help_for_argparse(help_text):
+        """Escape literal '%' while preserving argparse mapping placeholders.
+
+        argparse interpolates help text with ``%`` formatting against a dict.
+        Keep valid ``%(name)s``-style placeholders as-is and escape everything
+        else so help text like date formats (for example, ``%Y-%m-%d``) doesn't
+        crash during parser construction.
+        """
+        if not isinstance(help_text, str) or '%' not in help_text:
+            return help_text
+
+        result = []
+        idx = 0
+        text_len = len(help_text)
+        while idx < text_len:
+            char = help_text[idx]
+            if char != '%':
+                result.append(char)
+                idx += 1
+                continue
+
+            # Keep already-escaped percent signs as-is.
+            if idx + 1 < text_len and help_text[idx + 1] == '%':
+                result.append('%%')
+                idx += 2
+                continue
+
+            # Preserve mapping placeholders, e.g. %(default)s.
+            if idx + 1 < text_len and help_text[idx + 1] == '(':
+                closing_paren = help_text.find(')', idx + 2)
+                if closing_paren != -1 and closing_paren + 1 < text_len:
+                    conversion_char = help_text[closing_paren + 1]
+                    if conversion_char.isalpha():
+                        result.append(help_text[idx:closing_paren + 2])
+                        idx = closing_paren + 2
+                        continue
+
+            # Any other '%' is literal and must be escaped for argparse.
+            result.append('%%')
+            idx += 1
+
+        return ''.join(result)
+
+    @staticmethod
     def create_global_parser(cli_ctx=None):
         global_parser = argparse.ArgumentParser(prog=cli_ctx.name, add_help=False)
         arg_group = global_parser.add_argument_group('global', 'Global Arguments')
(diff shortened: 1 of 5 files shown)


bebound (dismissed):

knack/parser.py
```diff
@@ -32,6 +32,50 @@
 
 class CLICommandParser(argparse.ArgumentParser):
 
+    @staticmethod
+    def _sanitize_help_for_argparse(help_text):
```
jiasli, line 36: This function is a little bit too complex. A regular expression can do the job. AI response:

> Yes — the whole `_sanitize_help_for_argparse` loop can be replaced by a single `re.sub` with a negative lookahead. The idea: match any `%` that is **not** followed by another `%` (already-escaped) and **not** followed by a `%(name)x` mapping placeholder, then replace it with `%%`.
> 
> ```python name=knack/parser.py
> import re
> 
> # Matches a single '%' that is NOT:
> #   - followed by another '%'  (already escaped: %%)
> #   - the start of a mapping placeholder like %(default)s, %(prog)r, etc.
> _LITERAL_PERCENT_RE = re.compile(r'%(?!%|\([^()]+\)[A-Za-z])')
> 
> @staticmethod
> def _sanitize_help_for_argparse(help_text):
>     if not isinstance(help_text, str) or '%' not in help_text:
>         return help_text
>     return _LITERAL_PERCENT_RE.sub('%%', help_text)
> ```
> 
> How the lookahead reads:
> - `%%` → the `%` is followed by `%`, lookahead fails the "not" condition → left alone (the second `%` is then consumed on its own iteration; since it's followed by whatever came after `%%`, it usually matches and… wait — see caveat below).
> - `%(default)s` → the `%` is followed by `(default)s`, matches `\([^()]+\)[A-Za-z]` → left alone.
> - `%Y`, `%`, `% `, `100%` → no match in lookahead → replaced with `%%`.
> 
> ### Important caveat about `%%`
> 
> A naive single-pass `re.sub` walks left-to-right and **doesn't re-scan** its own replacement, but it does advance past the matched `%` only. With the lookahead `(?!%)`, the first `%` in `%%` is skipped, then the regex engine moves to the second `%`. That second `%` is followed by whatever comes next (e.g. end of string, a space, `Y`, etc.), so it would get doubled into `%%`, turning `%%` into `%%%`. ❌
> 
> To handle this correctly with one regex, match the `%%` pair explicitly and keep it, e.g. using an alternation + a callback:
> 
> ```python name=knack/parser.py
> _PERCENT_RE = re.compile(r'%%|%\([^()]+\)[A-Za-z]|%')
> 
> def _sanitize_help_for_argparse(help_text):
>     if not isinstance(help_text, str) or '%' not in help_text:
>         return help_text
>     # Keep %% and %(name)x as-is; escape any other lone %.
>     return _PERCENT_RE.sub(lambda m: m.group(0) if len(m.group(0)) > 1 else '%%', help_text)
> ```
> 
> This is equivalent in behavior to the hand-written loop in the PR but ~6 lines instead of ~40, and arguably easier to audit. The alternation order matters: `%%` and `%(...)x` are tried before the bare `%` fallback, so they "win" and are returned unchanged; only the lone `%` branch gets escaped.
> 
> If you'd like, I can open a follow-up PR replacing the loop with this regex version (plus keeping the existing tests, which already cover `%(default)s`, `100%%`, and `%Y-%m-%d`).

Test result:

```py
import re

_PERCENT_RE = re.compile(r'%%|%\([^()]+\)[A-Za-z]|%')

def _sanitize_help_for_argparse(help_text):
    if not isinstance(help_text, str) or '%' not in help_text:
        return help_text
    # Keep %% and %(name)x as-is; escape any other lone %.
    return _PERCENT_RE.sub(lambda m: m.group(0) if len(m.group(0)) > 1 else '%%', help_text)

print(_sanitize_help_for_argparse("A string with single %, double %%, and placeholders %(prog)s. Test test."))
```

Output:

```
A string with single %%, double %%, and placeholders %(prog)s. Test test.
```

jiasli: However, even this level of complexity is not necessary, as pointed out by https://github.com/microsoft/knack/pull/299#discussion_r3128791113, the sanitization can be achieved as below:

1. If there is `%(` or `%%` in the help string, leave the string as is, because this indicates the help string author is aware of the escaping rule.
3. If not, escape `%` to `%%`, like AWS CLI did (https://github.com/aws/aws-cli/pull/9790).

It is not necessary to handle `"a string with both % and %%"`. Just let it fail!

jiasli: AI comments on the above approach :

> Yes, the comment is valid — and arguably the best approach for this PR. It trades a bit of edge-case coverage for a large reduction in complexity, with a clear and defensible rule.
> 
> ### The proposed rule
> 1. If the help string contains `%(` or `%%`, the author already knows about argparse's escaping → leave it alone.
> 2. Otherwise, blindly escape every `%` → `%%`.
> 
> ### Why it's valid
> 
> **It's a "signal of intent" heuristic.** The presence of `%(` or `%%` is a strong indicator that the author wrote argparse-aware help. If they didn't use either, they almost certainly meant `%` literally (`%Y-%m-%d`, `50% off`, `ratio 100%`, etc.).
> 
> **Implementation collapses to ~5 lines, no regex needed:**
> 
> ```python name=knack/parser.py
> @staticmethod
> def _sanitize_help_for_argparse(help_text):
>     if not isinstance(help_text, str) or '%' not in help_text:
>         return help_text
>     if '%(' in help_text or '%%' in help_text:
>         return help_text  # author handled escaping themselves
>     return help_text.replace('%', '%%')
> ```
> 
> **Covers all the realistic cases in the PR's tests:**
> | Input | Has `%(`/`%%`? | Output | argparse result |
> |---|---|---|---|
> | `default=%(default)s prog=%(prog)s` | yes | unchanged | ✅ expands |
> | `ratio 100%%` | yes | unchanged | ✅ renders `100%` |
> | `bad % token` | no | `bad %% token` | ✅ renders `bad % token` |
> | `Expected format: %Y-%m-%d` | no | `Expected format: %%Y-%%m-%%d` | ✅ renders literally |
> 
> ### The conscious trade-off
> 
> Strings that mix both — e.g. `"Use %Y-%m-%d, default %(default)s"` — fall through rule 1 and crash argparse. The comment explicitly says **"just let it fail!"** That's reasonable because:
> - Such mixing is extremely rare in practice.
> - The failure is loud at parser-construction time, not silent corruption.
> - The author can fix it by escaping (`%%Y-%%m-%%d`).
> 
> ### How it compares to the regex approaches we discussed
> 
> | Approach | LOC | Handles `%Y` + `%(default)s` mixed | Cognitive load |
> |---|---|---|---|
> | Hand-written loop (current PR) | ~40 | yes | medium |
> | Alternation regex + lambda | ~5 | yes | medium |
> | Lookbehind/lookahead regex | ~5 | partial (`%%%` edge) | high |
> | **`%(`/`%%` heuristic (this comment)** | ~5 | no (by design) | **low** |
> 
> So the comment is valid both technically and as a design judgment. I'd recommend adopting it, and updating the `n8` test (`'bad % token'`) to assert the escaped-rendering path, plus dropping the `n7`/`%%` test's expectation since `%%` strings are now passed through untouched (which still renders as `%` via argparse — same observable result).

tests/test_parser.py
```diff
@@ -180,6 +180,22 @@ def remove_test_file(file):
...
+        command.add_argument('date_fmt', '--date-fmt', help='Expected format: %Y-%m-%d')
+        cmd_table = {'test command': command}
+        self.mock_ctx.commands_loader.command_table = cmd_table
+
+        parser = CLICommandParser()
+        parser.load_command_table(self.mock_ctx.commands_loader)
+
+    def test_help_string_preserves_argparse_placeholders(self):
+        sanitized = CLICommandParser._sanitize_help_for_argparse('default is %(default)s (100% expected)')
+        self.assertEqual(sanitized, 'default is %(default)s (100% expected)')
```
jiasli:
```suggestion
        sanitized = CLICommandParser._sanitize_help_for_argparse('default is %(default)s (100%% expected)')
        self.assertEqual(sanitized, 'default is %(default)s (100%% expected)')
```

jiasli:
Could you edit this PR to update the [history notes](https://github.com/microsoft/knack/blob/dev/HISTORY.rst)? 
 
Knack's version has already been bumped to 0.14 (https://github.com/microsoft/knack/pull/297), so there is no chance to update the history notes.

bebound (approved):

jiasli (approved):

